PR #1, opened 2022-07-21
Split programming model out of worker (library side)

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/568

Here's the PR for the worker side: https://github.com/Azure/azure-functions-nodejs-worker/pull/608

The most important file to look at is `src/InvocationModel.ts`


package.json
```diff
@@ -1,48 +1,77 @@
 {
-    "name": "azure-functions-nodejs-worker",
-    "author": "Microsoft Corporation",
-    "version": "3.4.0",
-    "description": "Microsoft Azure Functions NodeJS Worker",
-    "license": "(MIT OR Apache-2.0)",
+    "name": "@azure/functions",
+    "version": "3.5.0-alpha.1",
```
hossam-nasr, line 3: Where is this version coming from?

package.json
```diff
@@ -1,48 +1,77 @@
...
+    "description": "Microsoft Azure Functions NodeJS Framework",
```
hossam-nasr, line 4: I thought we were calling it "Library" not "Framework"?

scripts/updateVersion.ts
```diff
@@ -39,35 +32,18 @@ function validateVersion() {
...
-        !typesPackageJsonVersion ||
-        !parsedVersion ||
-        !parsedTypesVersion
-    ) {
+
+    if (!packageJsonVersion || !constantsVersion || !parsedVersion) {
         throw new Error('Failed to detect valid versions in all expected files');
-    } else if (nuspecVersion !== packageJsonVersion || constantsVersion !== packageJsonVersion) {
+    } else if (constantsVersion !== packageJsonVersion) {
         throw new Error(`Worker versions do not match.`);
```
hossam-nasr: "Worker" versions?

src/InvocationModel.ts
```diff
@@ -0,0 +1,173 @@
...
+
+import { AzureFunction, Context } from '@azure/functions';
+import * as coreTypes from '@azure/functions-core';
+import {
+    CoreInvocationContext,
+    InvocationArguments,
+    RpcInvocationResponse,
+    RpcLog,
+    RpcParameterBinding,
+} from '@azure/functions-core';
```
hossam-nasr, line 12: These imports are weird to me.

When importing `AzureFunction` and `Context` from `@azure/functions`, are we importing them from the current npm package (i.e. the worker), or are we importing them from ourselves?

Also, why are we importing `* as coreTypes` from `@azure/functions-core`, while excluding `CoreInvocationContext`, `InvocationArguments`, etc., and importing them on their own?

src/InvocationModel.ts
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    async getResponse(context: Context, result: unknown): Promise<RpcInvocationResponse> {
+        const response: RpcInvocationResponse = { invocationId: this.#coreCtx.invocationId };
+        response.outputData = [];
+        const info = this.#funcInfo;
+
+        // Allow HTTP response from context.res if HTTP response is not defined from the context.bindings object
+        if (info.httpOutputName && context.res && context.bindings[info.httpOutputName] === undefined) {
```
hossam-nasr, line 82: I'm assuming this is just copying old code, but just a javascript question here: when you type this condition like this, does this check that each of `info.httpOutputName`, `context,res`, etc., are each `=== undefined`, or does this first evaluate `info.httpOutputName` as a truth value (i.e., `info.httpOutputName !== undefined`), then does the same for `context.res`, etc.?

src/index.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as coreTypes from '@azure/functions-core';
+import { CoreInvocationContext, setProgrammingModel } from '@azure/functions-core';
```
hossam-nasr, line 5: Same question here about imports

src/index.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as coreTypes from '@azure/functions-core';
+import { CoreInvocationContext, setProgrammingModel } from '@azure/functions-core';
+import { version } from './constants';
+import { InvocationModel } from './InvocationModel';
+
+class ProgrammingModel implements coreTypes.ProgrammingModel {
+    name = '@azure/functions';
```
hossam-nasr, line 10: Is the name of the programming model the same as the name of the npm package?

src/Context.ts
```diff
@@ -22,20 +22,19 @@ import { FunctionInfo } from './FunctionInfo';
...
 import EventEmitter = require('events');
-import LogLevel = rpc.RpcLog.Level;
 
 export function CreateContextAndInputs(
     info: FunctionInfo,
-    request: rpc.IInvocationRequest,
-    userLogCallback: UserLogCallback
+    request: RpcInvocationRequest,
+    userLogCallback: UserLogCallback,
+    doneEmitter: EventEmitter
```
hossam-nasr, line 30: For the most part, this function seem to be copy-pasted from the old worker. The only difference I see here is that, before, `doneEmitter` was created by this function, whereas now it's passed to it. Curious what difference it makes and why this decision was made?

test/setupTestCoreApi.ts
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import Module = require('module');
+
+export function setupTestCoreApi(): void {
```
hossam-nasr, line 6: Not sure if I'm following this correctly, but it seems like here we're defining our own mock core API for testing purposes, which isn't necessarily the same core API provided by the worker. This means that if, for whatever reason, this API changes from the worker, we have to update it here too. I don't have ideas off the top of my head, but is there any way we can test against the actual core API directly, or link the two somehow?

hossam-nasr (approved):
In general, lgtm :) I left a few nits/questions, but nothing I wanna block over. Also assuming that most of the code is copied over from the old worker, while modifying types to use types provided by the core API, so didn't spend too much analyzing the details of what the invocation model does

package.json
ejizba: First of all I incremented the previous version from "3.4.0" to "3.5.0". Then I added "-alpha.1" because I want to use a "preview" version for a while before giving it a "production ready" version. We are doing a big transition from a types packages to a "types & source" package and I want to make sure we don't break people

package.json
ejizba: The only reason I went with library was so the repo name would be consistent with other languages. I still think "framework" is the correct word, so that's what I plan to use for any display string. That being said, we can change display strings at any time, so it's not a hard and fast decision either way

src/InvocationModel.ts
ejizba: > When importing `AzureFunction` and `Context` from `@azure/functions`, are we importing them from the current npm package (i.e. the worker), or are we importing them from ourselves?

I'm a bit confused by this question. The "current npm package" is not the worker, it's the library. And it really just translates to the `types/index.d.ts` file in this repo (relevant line in tsconfig [here](https://github.com/Azure/azure-functions-nodejs-library/blob/v3.x/tsconfig.json#L12)). We import two-ish things from "ourselves" - for things that go away after you compile to JavaScript (think `interface` and `type`) we will import from the types file in this repo (aka `@azure/functions`). For other things that are real source code and actually stay when compiled to javascript (`class`, `const`, `function` etc.) we will import from the `src` folder directly

> Also, why are we importing `* as coreTypes` from `@azure/functions-core`, while excluding `CoreInvocationContext`, `InvocationArguments`, etc., and importing them on their own?

This is easier to answer. I prefer "importing them on their own". However, I use `* as coreTypes` for lines like this where the class name and interface name are the same and would conflict with each other:
```
export class InvocationModel implements coreTypes.InvocationModel {
```

src/InvocationModel.ts
ejizba: Yeah just copying code.

First of all, `info.httpOutputName` doesn't just check `=== undefined`. The condition will be false for any false-y value like `undefined`, `null`, `''`, `0`, `false` and maybe more. Sometimes this can lead to unexpected behavior and weird bugs, but we have lint rules in place to help with that.

Secondly, if the first condition returns false, it will not check the following conditions for an `&&` condition. So if `info.httpOutputName` is false, it will not check `context.res`.

I think that's what you were asking...

src/index.ts
ejizba: Yeah, although it doesn't really matter. I'm just using the name here for tracking purposes in telemetry

src/Context.ts
ejizba: I needed to share `doneEmitter` between the InvocationModel methods `getArguments`, `invokeFunction`, and `getResponse` and it was easiest to create it on the class and pass it in here. Suffice to say it was just a byproduct of how the new code is structured and not some sort of grand design decision

test/setupTestCoreApi.ts
ejizba: First of all, we don't have to mock the entire core api, just a few trivial enums. So I'm not worried about the engineering upkeep of the api changing

And yes there is a way to link the two - you'd have to essentially run the `func` cli as a part of the test. However that's more of an "end-to-end" test and even if we do that (which I want to do, just not sure when) it would not replace these unit tests. End-to-end tests take a lot longer to run and usually integrate external components. Unit tests should be super fast/simple and not rely on external components

scripts/updateVersion.ts
ejizba: Fixed. Fwiw, I expect we will be removing references to the word "worker" for a while in this repo.

PR #2, opened 2022-08-18
Update readme and notice

ejizba:


ejizba:
feel free to review, but merging to unblock https://github.com/Azure/azure-functions-nodejs-worker/pull/608

PR #4, opened 2022-08-19
Add bufferBody property to http request

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/294, we want to add a property on the http request that provides the body as a buffer. In order to do that we need to:
1. Turn on the `RawHttpBodyBytes` capability in the worker. This tells the host to send us the body in bytes, never a string. (NOTE: We don't want bytes _and_ string because that duplicates the data sent over rpc and the data can be large)
2. Since the host no longer converts it to string for us, we have to replicate the previous host behavior for existing properties (`body` and `rawBody`) so that we don't break people

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
             "license": "MIT",
             "dependencies": {
                 "fs-extra": "^10.0.1",
+                "iconv-lite": "^0.6.3",
                 "long": "^4.0.0",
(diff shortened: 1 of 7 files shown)


hossam-nasr (approved):
Approved with the caveat that all the decoding/encoding stuff completely flies over my head, but I trust you did your due diligence😛

PR #10, opened 2022-08-25
Add HttpRequest.get method to types

ejizba:
It's the recommended way to get a header because it works regardless of casing

Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/274

### DIFF
--- a/types/http.d.ts
+++ b/types/http.d.ts
@@ -109,6 +109,11 @@ export interface HttpRequest {
     bufferBody?: Buffer;
 
     /**
+     * Get the value of a particular header field
+     */
+    get(field: string): string | undefined;
+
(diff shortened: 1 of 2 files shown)


types/http.d.ts
```diff
@@ -108,6 +108,11 @@ export interface HttpRequest {
      */
     bufferBody?: Buffer;
 
+    /**
+     * Get the value of a particular header field
+     */
+    get(field: string): string | undefined;
```
hossam-nasr, line 114: So this property was already there just not explicitly there in the types?

ejizba: Yes. In general, the types are pretty lacking

PR #11, opened 2022-08-25
Clean up error naming

ejizba:
Same as https://github.com/Azure/azure-functions-nodejs-worker/pull/624


hossam-nasr (commented):
I don't see where in the library code the property `isAzureFunctionsSystemError` is being used? Or is are the other supposed to be catched by the worker, which _does_ care?

ejizba:
It'll be caught by the worker. That's why I wanted to do this for the "first" release with the repos split up because it's used across repos and we won't want to change it moving forward. That also relates to why it's such a long descriptive name, because we don't want users accidentally setting the property for their own reasons (Like...if we called it "isInternalError" or "isSystemError" you could see other people re-using that property name for something else)

PR #14, opened 2022-09-02
Make RPC HTTP converter copy input by value

### LINKED ISSUE #13: Multiple calls to `context.res.end()` affect each other
From [CRI](https://portal.microsofticm.com/imp/v3/incidents/details/318403184/home)

In the `HttpResponseFull` object `context.res` that is provided to HTTP functions, the `context.res.end()` method does two things: First, it sets the `body` property to whatever argument is provided, then they call `context.done()`. 

If `context.res.end()` is called more than once, even though subsequent calls to `context.done()` should be ignored, the `response.body` property is still modified, since the RPC converters copy the input by reference not by value. This means that subsequent calls to `context.res.end()` will still modify the `.body` property of the HTTP output in the invocation response, even though semantically it should not, and also bypassing the RPC conversions, causing a crash.

Even though multiple calls to `context.done()` are not supported, this should still be fixed because:

1. Semantically, subsequent calls should not have any effect
2. It is dangerous to allow our response to the host to be modified after passing our checks
3. Such an error shouldn't completely break customers

Therefore, RPC converters should make sure that they copy all provided arguments by value.

#### Repro steps

1. Create an HTTP triggered function that calls `context.res.end()` twice, once without an argument and once with a `string` argument:

```TS
const httpTrigger: AzureFunction = function (context: Context, req: HttpRequest): void {
    context.log('HTTP trigger function processed a request.');
    context.res.end();
    context.res.end('test test')
};
```
2. Trigger the function
3. You see an error like this:

```
System.Private.CoreLib: Exception while executing function: Functions.index. System.Private.CoreLib: node exited with code 1
     at Object.callErrorFromStatus (C:\Users\hossamnasr\ms\azure\nodejs-worker\azure-functions-nodejs-worker\node_modules\@grpc\grpc-js\build\src\call.js:31:26),  details: 'Request message serialization failure: .AzureFunctionsRpcMessages.RpcHttp.body: object expected',,LanguageWorkerConsoleLog[error] Worker c4d2aa66-b3c5-4e81-90c1-dfbe44b82f92 uncaught exception: Error: Error: 13 INTERNAL: Request message serialization failure: .AzureFunctionsRpcMessages.RpcHttp.body: object expected     at ClientDuplexStreamImpl.<anonymous> (C:\Users\hossamnasr\ms\azure\nodejs-worker\azure-functions-nodejs-worker\dist\src\setupEventStream.js:38:15)     at ClientDuplexStreamImpl.emit (node:events:390:28)     at Object.onReceiveStatus (C:\Users\hossamnasr\ms\azure\nodejs-worker\azure-functions-nodejs-worker\node_modules\@grpc\grpc-js\build\src\client.js:390:28)     at Object.onReceiveStatus (C:\Users\hossamnasr\ms\azure\nodejs-worker\azure-functions-nodejs-worker\node_modules\@grpc\grpc-js\build\src\client-interceptors.js:299:181)     at C:\Users\hossamnasr\ms\azure\nodejs-worker\azure-functions-nodejs-worker\node_modules\@grpc\grpc-js\build\src\call-stream.js:145:78     at processTicksAndRejections (node:internal/process/task_queues:78:11).
```

#### Expected behavior

* Function triggers without an error
* The second call to `context.res.end()` is ignored

#### Actual behavior

* Function throws the error above
* The second call to `context.res.end()` modified the `response.body` field _after_ RPC conversions are made, and RPC conversions aren't initiated again, leading the actual invocation response object to contain an RPC HTTP output object with incompatible types, throwing the above error.

#### Known workarounds

* Don't use multiple `context.done()` calls, which is unsupported anyway
* Pin to Host version `3.5.2.0`

hossam-nasr:
Fixes #13. This PR modifies the `toRpcHttp` function such that it copies the provided HTTP response object rather than modifying it in-place, and therefore preventing future alterations to the input from affecting the output of the function. Also enforces stricter types on the input.

### DIFF
--- a/src/converters/RpcHttpConverters.ts
+++ b/src/converters/RpcHttpConverters.ts
@@ -1,7 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { Cookie } from '@azure/functions';
+import { Cookie, HttpResponse } from '@azure/functions';
 import {
     RpcHttpCookie,
     RpcHttpCookieSameSite,
@@ -55,7 +55,7 @@ export function fromNullableMapping(
  * 'http' types are a special case from other 'ITypedData' types, which come from primitive types.
  * @param inputMessage  An HTTP response object
  */
-export function toRpcHttp(inputMessage): RpcTypedData {
+export function toRpcHttp(inputMessage: HttpResponse): RpcTypedData {
     // Check if we will fail to find any of these
     if (typeof inputMessage !== 'object' || Array.isArray(inputMessage)) {
         throw new AzFuncSystemError(
@@ -63,15 +63,18 @@ export function toRpcHttp(inputMessage): RpcTypedData {
         );
     }
 
-    const httpMessage: RpcHttpData = inputMessage;
-    httpMessage.headers = toRpcHttpHeaders(inputMessage.headers);
-    httpMessage.cookies = toRpcHttpCookieList(inputMessage.cookies || []);
     let status = inputMessage.statusCode;
     if (typeof inputMessage.status !== 'function') {
         status ||= inputMessage.status;
     }
-    httpMessage.statusCode = status && status.toString();
-    httpMessage.body = toTypedData(inputMessage.body);
+    const httpMessage: RpcHttpData = {
+        ...inputMessage,
+        statusCode: status?.toString() || null,
+        headers: toRpcHttpHeaders(inputMessage.headers),
+        cookies: toRpcHttpCookieList(inputMessage.cookies || []),
+        body: toTypedData(inputMessage.body),
+    };
+
     return { http: httpMessage };
 }
 
@@ -79,11 +82,13 @@ export function toRpcHttp(inputMessage): RpcTypedData {
  * Convert HTTP headers to a string/string mapping.
  * @param inputHeaders
  */
-function toRpcHttpHeaders(inputHeaders: RpcTypedData) {
+function toRpcHttpHeaders(inputHeaders: RpcTypedData | undefined) {
     const rpcHttpHeaders: { [key: string]: string } = {};
-    for (const key in inputHeaders) {
-        if (inputHeaders[key] != null) {
-            rpcHttpHeaders[key] = inputHeaders[key].toString();
+    if (inputHeaders) {
+        for (const key in inputHeaders) {
+            if (inputHeaders[key] != null) {
+                rpcHttpHeaders[key] = inputHeaders[key].toString();
+            }
         }
     }
     return rpcHttpHeaders;
--- a/src/converters/RpcHttpConverters.ts
+++ b/src/converters/RpcHttpConverters.ts
@@ -55,14 +55,15 @@ export function fromNullableMapping(
  * 'http' types are a special case from other 'ITypedData' types, which come from primitive types.
  * @param inputMessage  An HTTP response object
  */
-export function toRpcHttp(inputMessage: HttpResponse): RpcTypedData {
+export function toRpcHttp(data: unknown): RpcTypedData {
     // Check if we will fail to find any of these
-    if (typeof inputMessage !== 'object' || Array.isArray(inputMessage)) {
+    if (typeof data !== 'object' || Array.isArray(data)) {
         throw new AzFuncSystemError(
             "The HTTP response must be an 'object' type that can include properties such as 'body', 'status', and 'headers'. Learn more: https://go.microsoft.com/fwlink/?linkid=2112563"
         );
     }
 
+    const inputMessage: HttpResponse = data || {};
     let status = inputMessage.statusCode;
     if (typeof inputMessage.status !== 'function') {
         status ||= inputMessage.status;
--- a/test/converters/BindingConverters.test.ts
+++ b/test/converters/BindingConverters.test.ts
@@ -7,12 +7,11 @@ import { fromString } from 'long';
 import 'mocha';
 import { getBindingDefinitions, getNormalizedBindingData } from '../../src/converters/BindingConverters';
 import { fromTypedData } from '../../src/converters/RpcConverters';
-import { toRpcHttp } from '../../src/converters/RpcHttpConverters';
 import { FunctionInfo } from '../../src/FunctionInfo';
 
 describe('Binding Converters', () => {
     it('normalizes binding trigger metadata for HTTP', () => {
-        const mockRequest: RpcTypedData = toRpcHttp({ url: 'https://mock' });
+        const mockRequest: RpcTypedData = { http: { url: 'https://mock' } };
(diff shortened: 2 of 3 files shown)


src/converters/RpcHttpConverters.ts
```diff
@@ -55,35 +55,40 @@ export function fromNullableMapping(
...
-    httpMessage.statusCode = status && status.toString();
-    httpMessage.body = toTypedData(inputMessage.body);
+    const httpMessage: RpcHttpData = {
+        ...inputMessage,
+        statusCode: status?.toString() || null,
+        headers: toRpcHttpHeaders(inputMessage.headers),
+        cookies: toRpcHttpCookieList(inputMessage.cookies || []),
+        body: toTypedData(inputMessage.body),
+    };
+
```
hossam-nasr, line 78: Most of these modifications were to make sure Typescript is happy after enforcing the `HttpResponse` type on the argument. The most crucial modfication though is the `...inputMessage` piece

src/converters/RpcHttpConverters.ts
```diff
@@ -55,35 +55,40 @@ export function fromNullableMapping(
...
+function toRpcHttpHeaders(inputHeaders: RpcTypedData | undefined) {
     const rpcHttpHeaders: { [key: string]: string } = {};
-    for (const key in inputHeaders) {
-        if (inputHeaders[key] != null) {
-            rpcHttpHeaders[key] = inputHeaders[key].toString();
+    if (inputHeaders) {
+        for (const key in inputHeaders) {
+            if (inputHeaders[key] != null) {
+                rpcHttpHeaders[key] = inputHeaders[key].toString();
+            }
```
hossam-nasr, line 92: This was also to make TypeScript happy because `HttpResponse.headers` could be undefined

test/converters/BindingConverters.test.ts
```diff
@@ -7,12 +7,11 @@ import { fromString } from 'long';
 import 'mocha';
 import { getBindingDefinitions, getNormalizedBindingData } from '../../src/converters/BindingConverters';
 import { fromTypedData } from '../../src/converters/RpcConverters';
-import { toRpcHttp } from '../../src/converters/RpcHttpConverters';
 import { FunctionInfo } from '../../src/FunctionInfo';
 
 describe('Binding Converters', () => {
     it('normalizes binding trigger metadata for HTTP', () => {
-        const mockRequest: RpcTypedData = toRpcHttp({ url: 'https://mock' });
+        const mockRequest: RpcTypedData = { http: { url: 'https://mock' } };
```
hossam-nasr, line 14: This was the only instance in the code where `toRpcHttp` was used to convert an HTTP request rather than an HTTP response. I changed it in favour of having stricter type enforcement

src/converters/RpcHttpConverters.ts
```diff
@@ -55,35 +55,40 @@ export function fromNullableMapping(
  * 'http' types are a special case from other 'ITypedData' types, which come from primitive types.
  * @param inputMessage  An HTTP response object
  */
-export function toRpcHttp(inputMessage): RpcTypedData {
+export function toRpcHttp(inputMessage: HttpResponse): RpcTypedData {
```
ejizba: The user could pass us literally anything as the inputMessage so I'd rather be a bit more careful and use `unknown` as the type until after we've validated it on line 60. (Before your change, the implicit type was `any` which is the worst possible type. `HttpResponse` is somewhat better, but `unknown` is best in this case) 

Something like this:
```typescript
export function toRpcHttp(data: unknown): RpcTypedData {
    // Check if we will fail to find any of these
    if (typeof data !== 'object' || Array.isArray(data)) {
        throw new AzFuncSystemError(
            "The HTTP response must be an 'object' type that can include properties such as 'body', 'status', and 'headers'. Learn more: https://go.microsoft.com/fwlink/?linkid=2112563"
        );
    }
    const inputMessage: HttpResponse = data || {};
```

ejizba (approved):

src/converters/RpcHttpConverters.ts
hossam-nasr: @ejizba As far as I understand, this function isn't called directly by user code, but rather our code calls it using user input. Isn't it better validation to enforce the `HttpResponse` type on the argument so that TypeScript throws an error if our code tries to pass user input directly without ensuring it's of an `HttpResponse` type? I'm saying this because the validation that this function does (ensuring it's an object) isn't enough to guarantee it's an `HttpResponse`.

ejizba: > Isn't it better validation to enforce the HttpResponse type on the argument so that TypeScript throws an error if our code tries to pass user input directly without ensuring it's of an HttpResponse type?

For TypeScript users this is true, but a lot of our users (probably the majority?) are JavaScript users so a TypeScript type doesn't enforce anything at all. That's why we need to do both.

> I'm saying this because the validation that this function does (ensuring it's an object) isn't enough to guarantee it's an HttpResponse.

Sure it might not be perfect, but TypeScript can do a lot of validation for us and we should leverage that whenever possible. None of the properties on `HttpResponse` are required, which is why TypeScript only requires us to validate that `data` is an object.

The v3.x branch uses `any` waaayyy too much which means we lose a lot of type validation. I've mostly fixed it on v4.x, but if everything was properly typed this wouldn't even be a discussion. You would have to do `data: unknown` otherwise you'd get a build error if you did `data: HttpResponse`. It will be an unhill battle to get rid of `any` on v3.x for historical reasons, but this is just one example where I would like us to do better.

hossam-nasr: >> Isn't it better validation to enforce the HttpResponse type on the argument so that TypeScript throws an error if our code tries to pass user input directly without ensuring it's of an HttpResponse type?

> For TypeScript users this is true, but a lot of our users (probably the majority?) are JavaScript users so a TypeScript type doesn't enforce anything at all. That's why we need to do both.

But our code is the only one that uses this function, and our code is only TypeScript, no?

ejizba: Our code is the only one that uses this function, yes. There are several layers up the stack that are using `any` and should be using `unknown`. I'm not asking you to fix the whole call stack because that would be a pain - but eventually you would get to a point where it's closer to user code

hossam-nasr: > The v3.x branch uses `any` waaayyy too much which means we lose a lot of type validation. I've mostly fixed it on v4.x, but if everything was properly typed this wouldn't even be a discussion

Are you saying that, because our code uses `any` so much, the fact that we don't get any build errors when saying `data: HttpResponse` doesn't guarantee that we are actually always passing an `HttpResponse` to this function, but that there is some validation missing, because `any` is used? Am I understanding this correctly?

ejizba: Yes there is some validation missing

hossam-nasr: That makes sense, thanks!

PR #25, opened 2022-09-27
Don't ignore return value of functions

hossam-nasr:
This PR makes it such that return values of functions are not ignored, and instead set as the `returnValue` property of the gRPC `functionInvocationResponse` message, even if no explicit output binding is set. This helps in cases such as Durable Functions where the function invocation is invoked programmatically and the return value is used without an explicit output binding

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,6 @@
             2
         ],
         "deprecation/deprecation": "error",
-        "@typescript/no-unsafe-assignment": "off",
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-namespace": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,10 @@
 {
     "parser": "@typescript-eslint/parser",
-    "plugins": ["@typescript-eslint", "header", "deprecation"],
+    "plugins": [
+        "@typescript-eslint",
+        "header",
+        "deprecation"
+    ],
     "parserOptions": {
         "project": "tsconfig.json",
         "sourceType": "module"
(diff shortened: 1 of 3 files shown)


src/InvocationModel.ts
```diff
@@ -104,6 +104,10 @@ export class InvocationModel implements coreTypes.InvocationModel {
             }
         }
 
+        if (!response.returnValue && response.outputData.length == 0) {
```
ejizba: I tested this on an http trigger with no output binding and it _is_ returning the value. It literally just shows the json as the response which seems pretty weird
Screen Shot 2022-10-07 at 2 14 08 PM

In other words, I do think we have to be more careful on this. Possible solutions include
1. Only set return value here based on trigger (either "not http" or "is durable"). This would match old behavior
2. Ask durable to pass in a flag somewhere saying "returnIfNoBindingSet". This would require changes in both repos
3. Ask durable to configure an output binding under the covers for the user. This would only require changes in durable

ejizba (changes requested):

src/InvocationModel.ts
hossam-nasr: I would favor the first option (specifically the "if not HTTP" part) since it matches old behavior anyway. Option 3 isn't really feasible because there isn't a valid output binding type that Durable can set

ejizba: I'm okay with the first option to unblock durable, although I would still prefer if we had a better reason than "it matches old behavior". The old behavior had a lot of weird things that did not make sense

hossam-nasr: Went with the first option to unblock Durable and added a detailed comment on why this is useful

.eslintrc.json
```diff
@@ -1,10 +1,6 @@
 {
```
ejizba: Please revert these unrelated changes, otherwise looks good

ejizba (approved):

PR #33, opened 2022-10-26
Add a flag to allow suppressing of bad pattern errors/warnings

### LINKED ISSUE #22: Add a flag to allow users to suppress the unrecommended `context.done()` patterns error log
This issue was inspired by this Durable SDK issue: https://github.com/Azure/azure-functions-durable-js/issues/373.

Some patterns related to use of `context.done()` are unsafe/unrecommended, and would throw a warning if used by user code, such as calling `context.done()` within an async function, or calling `context.done()` twice. However, there was a bug in the Node worker before, which suppressed this error message from being logged. This bug was fixed in https://github.com/Azure/azure-functions-nodejs-worker/pull/555, which meant this error message was now being shown. However, since this error message was suppressed before, some users who weren't getting an error message before, are not getting it, leaving with the option to either live with the error message or do a potentially breaking change. One such example is the Durable SDK, see discussion here for more details: https://github.com/Azure/azure-functions-durable-js/pull/380#discussion_r960099812.

This issue is to add a flag in the `context.done()` method to allow users to indicate they know what they are doing and to suppress the error message

hossam-nasr:
Resolves #22 and unblocks https://github.com/Azure/azure-functions-durable-js/issues/373. See issues for more details


src/InvocationModel.ts
```diff
@@ -151,8 +151,8 @@ export class InvocationModel implements coreTypes.InvocationModel {
         this.#log(level, 'system', ...args);
     }
 
-    #userLog(level: RpcLogLevel, ...args: any[]): void {
-        if (this.#isDone && this.#coreCtx.state !== 'postInvocationHooks') {
+    #userLog(level: RpcLogLevel, context: Context, ...args: any[]): void {
+        if (this.#isDone && this.#coreCtx.state !== 'postInvocationHooks' && !context.suppressBadPatternError) {
```
ejizba: This one was not mentioned in the issue and is only a warning. It also happens in different (although related) circumstances. I don't think the flag should hide it

On a more generic note, this flag should only be for one specific error, not a group of patterns. We have no incentive to make it easy to do this - all of these errors are valid and should be fixed in the vast majority of cases

types/Context.d.ts
```diff
@@ -62,6 +62,12 @@ export interface Context {
     res?: {
         [key: string]: any;
     };
+    /**
+     * If this flag is set to true in your function, error/warning logs for unrecommended `context.done()` patterns
+     * (e.g., using `context.done()` with `async`/`await`, calling `context.done()` twice) are suppressed.
+     * @default false
+     */
+    suppressBadPatternError?: boolean;
```
ejizba: `suppressBadPatternError` is too generic. How about `suppressAsyncDoneError` (which also matches existing terminology like `asyncDoneLearnMoreLink`)

types/Context.d.ts
```diff
@@ -62,6 +62,12 @@ export interface Context {
     res?: {
         [key: string]: any;
     };
+    /**
+     * If this flag is set to true in your function, error/warning logs for unrecommended `context.done()` patterns
```
ejizba: This description should link to the same `asyncDoneLearnMoreLink` used in the code

test/InvocationModel.test.ts
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { CoreInvocationContext, RpcInvocationRequest, RpcParameterBinding } from '@azure/functions-core';
```
ejizba: The src vs test code is out of balance again so I would like you to remove this file. Last time I did not block the PR because usually people have the opposite problem (not enough testing), but this time I think I will. It's not an efficient use of your time to author and my time to review, so I don't want this to keep happening

We already have tests in the worker covering this scenario [here](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/test/eventHandlers/InvocationHandler.test.ts#L535). You _could_ add a test there to reflect your new flag, but honestly my preference is that you don't. Your src code changes amounted to a single boolean in one if-condition and the test itself would be more complicated than that

types/Context.d.ts
hossam-nasr: That would make sense but the `suppressAsyncDoneError` would not cover the calling `context.done()` twice error. Or do we want this to suppress only the error in case of calling `context.done()` in an async function?

src/InvocationModel.ts
hossam-nasr: I was thinking of this as more of a "let me do what I want, I know what I'm doing" kind of flag. If a user has some use case where they think it's valid to call `context.done()` in an async function, they may also, for the same reason, want to log after calling `context.done()`

ejizba: They may also think that, but they should have to set another flag to suppress it. The user should have to explicitly choose to ignore each and every warning - we have those warnings for a good reason

types/Context.d.ts
ejizba: I think it would cover it. It mentions "done", which can cover the multiple done case. We don't have to be as explicit as `suppressAsyncDoneOrDoneDoneError` lol. I just don't want a generic name. If you have other "specific" ideas feel free to suggest them - `suppressExtraDoneError` maybe?

hossam-nasr: How does `suppressDoneErrors` sound?

src/InvocationModel.ts
hossam-nasr: Sure. Should I also add that second flag, or wait to see if users ask for it?

test/InvocationModel.test.ts
hossam-nasr: Yeah I didn't realize those same tests existed on the worker repo. I created this file more as a way of testing the existing code (and providing a scaffolding for future `InvocationModel` tests that I thought didn't exist yet), than testing my own changes.

That said, if the code lives here, why are the tests in the worker repo? Those tests won't even make sense anymore if at some point we decide to switch the default framework to v4

src/InvocationModel.ts
ejizba: Wait to see if people ask for it. Even if they ask, we will need a compelling reason (in Durable's case, the compelling reason is because they can't fix for backwards compat reasons)

hossam-nasr: Fair 👍🏾

types/Context.d.ts
ejizba: `suppressDoneErrors` still feels like you're going to suppress a bunch of different errors at once. Maybe I do want you to _only_ suppress the asyncDone error (aka not the `context.done` twice error). That's the only one affecting durable right?

src/InvocationModel.ts
hossam-nasr: Removed this

types/Context.d.ts
hossam-nasr: That is correct, that is the one affecting Durable

test/InvocationModel.test.ts
ejizba: > That said, if the code lives here, why are the tests in the worker repo?

First of all, this whole library pattern is still a baby okay 😝 It will be a while before everything is put in the "right" place and you'll just have to look in both places for a while. Secondly, I wanted to keep a decent amount of testing in the worker to ensure the worker & library are integrating correctly. Eventually we will want some proper "integration" testing between the two, but that doesn't really exist yet. Lastly, I'm not worried about defaulting to v4 - that's likely years away.

Your heart is in the right place - I love when people improve the test/engineering infrastructure. It's just not sustainable if you try to do that as a part of every bug fix and feature. You have to do it separately and consciously choose which engineering work is most important. Just because you happen to touch this code, doesn't mean this code is most in need of test/engineering improvement.

types/Context.d.ts
hossam-nasr: Changed it so that this only suppresses the one error affecting durable

types/Context.d.ts
hossam-nasr: Added link

test/InvocationModel.test.ts
hossam-nasr: Ok, removed this test file

ejizba (approved):

PR #39, opened 2022-10-30
Removing `methods` property

### LINKED ISSUE #38: When providing `HttpFunctionOptions` to the shortcut methods of `get`/`post`/etc., `method` shouldn't be settable
It's possible to write this:

```ts
app.get("get-user-by-slug", {
  route: "users/{slug}",
  handler: getUserBySlug,
  methods: ["PUT"],
});
```

Setting the `methods` doesn't _do anything_ (as it shouldn't):

```
get-user-by-slug: [GET] http://localhost:7071/api/users/{slug}
```

We should omit it from the type signature for the shortcut methods.

aaronpowell:
Addressing #38 

The `methods` property is now omitted from the type signature of `HttpFunctionOptions` if you're using the `get`/`post`/etc. shortcut methods to define a HTTP Trigger Function.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,6 +29,7 @@ import {
     HttpFunctionOptions,
     HttpHandler,
     HttpMethod,
+    HttpMethodFunctionOptions,
     HttpOutput,
     HttpOutputOptions,
     HttpTrigger,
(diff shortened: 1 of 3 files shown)


ejizba (approved):
Thanks!

PR #44, opened 2022-12-05
Replace our form support with undici's [v4.x]

ejizba:
Undici added support for forms in a recent release, so we can get rid of a bunch of our own code 🎉
- Release notes: https://github.com/nodejs/undici/releases/tag/v5.11.0
- PR: https://github.com/nodejs/undici/pull/1606

I copied over some of the old tests just to make sure it's generally working and it is. There are some differences in behavior, but since this is only for v4.x I'm not worried about breaking changes. I'm actually really glad undici landed form support before we announced official v4.x stuff.


test/http/HttpRequest.test.ts
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
hossam-nasr, line 1: Oh no, looks like your tests here are a lot longer than your actual code changes 🤯

hossam-nasr (approved):
LGTM!! Great that undici is doing this before we go GA, really simplifies our code :)

test/http/HttpRequest.test.ts
ejizba: Yes there are some nuances to that rule

First, this PR deletes a net of 270 lines of test code (from 380 -> 110). I _could_ have kept basically all the tests from before, but I thought that would have been out of balance which is why I deleted the majority.

Second, when switching from one implementation to another, it's always good to keep some of the old tests to make sure you have a decent level of continuity.

Lastly, these tests fit into a category of integration testing where _our_ code isn't complex, but an underlying dependency has very complex code which is worth testing. You'll notice each test is a pretty basic scenario, and I haven't really left any tests for edge cases.

hossam-nasr: That makes a lot of sense, thank you!

PR #45, opened 2022-12-07
Use undici request properties more directly

ejizba:
Similar goal as https://github.com/Azure/azure-functions-nodejs-library/pull/44, but much smaller scope. I just want to be more consistent with undici (which uses `readonly` on basically all class properties) and directly reference those properties instead of referencing the rpc ones in our constructor.

### DIFF
--- a/types/http.d.ts
+++ b/types/http.d.ts
@@ -79,34 +79,34 @@ export declare class HttpRequest {
     /**
      * HTTP request method used to invoke this function.
      */
-    method: HttpMethod;
+    readonly method: string;
 
     /**
      * Request URL.
      */
-    url: string;
+    readonly url: string;
 
     /**
      * HTTP request headers.
      */
-    headers: Headers;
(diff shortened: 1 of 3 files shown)


types/http.d.ts
```diff
@@ -79,34 +79,34 @@ export declare class HttpRequest {
     /**
      * HTTP request method used to invoke this function.
      */
-    method: HttpMethod;
+    readonly method: string;
```
hossam-nasr, line 82: Is it not a downgrade in types for this to be just any string? Does `undici` not enforce its own type on the method?

ejizba: `undici` uses string and I generally trust them to know more about this than us. My hunch is that our previous types were technically incorrect for some reason, for example casing or the fact that there are actually a bunch of possible methods (https://webconcepts.info/concepts/http-method/).

hossam-nasr (approved):

PR #46, opened 2022-12-09
Add `HttpResponse` class and model it after undici

### LINKED ISSUE #36: Provide a `Response` class for HTTP trigger functions
With a HTTP Trigger callback, we return an object with all the information for the response, from the body to status code, headers, etc.

This can feel a bit cumbersome, especially if you're coming from other API frameworks, such as Express or Fastify, where you have a `response` parameter and the function only returns the `body`.

Doing so would mean you could simplify code such as this:

```ts
app.get('route', (_, req) => {
  const headers: { [key of string]: string } = {}

  if (req.headers.has('accepts')) {
    const accepts = req.headers.get('accepts');
    if (accepts.indexOf('application/json')) {
      headers['content-type'] = 'application/json';
    }
  }

  return { headers, body: { foo: 'bar' } };
});
```

To:

```ts
app.get('route', (_, req, res) => {
  if (req.headers.has('accepts')) {
    const accepts = req.headers.get('accepts');
    if (accepts.indexOf('application/json')) {
      res.headers.add('content-type', 'application/json');
    }
  }

  return { foo: 'bar' };
});
```

Admittedly, this is the kind of thing that could be provided by a "higher level" wrapper around the low-level Functions API.

ejizba:
## new class
Here's the current way of returning a response:
```typescript
return { body: `Hello, ${name}!` };
```

In addition to that, you can now use a new `HttpResponse` class:
```typescript
const res = new HttpResponse({ body: `Hello, ${name}!` });
res.headers.set('headerName', 'headerValue');
return res;
```

## using undici's body type

When creating the `HttpResponse` class, I modeled it directly off of undici. The biggest change was around the body type, where I completely ditched ours for undici's. Under the covers, all responses get translated to a buffer by undici before we send the data over rpc. This actually makes our treatment of requests and responses more consistent because all requests come in as a buffer from the host and we let undici handle translating it to string/buffer/form/etc..

Old: 
```typescript
export type HttpResponseBody = string | Buffer | NodeJS.ArrayBufferView | number | object;
```
New (we're directly using undici's type):
```typescript
export type BodyInit =
  | ArrayBuffer
  | AsyncIterable<Uint8Array>
  | Blob
  | FormData
  | Iterable<Uint8Array>
  | NodeJS.ArrayBufferView
  | URLSearchParams
  | null
  | string
```

## json body

Within the BodyInit type, undici has a lot of nice new types we didn't support but there are also a few missing from our old stuff. First, `number` which I don't care about. Second, `object` which is how we handled json which I do care about. I assume undici doesn't list `object` because it would conflict with several of their other types and generally make things ambiguous. Undici supports json bodies through a static `Response.json` method which creates a Response, serializes the body, and adds the `application/json` content-type header. I'm using that method under the covers, but I didn't actually add a similar static method to our class. Instead, I chose to add a `jsonBody` property to our `HttpResponseInit` type so that people could still do json with or without the class. For example:

```typescript
return { jsonBody: { foo: 'bar' } };
return new HttpResponse({ jsonBody: { foo: 'bar' } });
```

Fixes #36

### DIFF
--- a/src/http/HttpResponse.ts
+++ b/src/http/HttpResponse.ts
@@ -1,0 +1,64 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as types from '@azure/functions';
+import { HttpResponseInit } from '@azure/functions';
+import { Blob } from 'buffer';
+import { ReadableStream } from 'stream/web';
+import { FormData, Headers, Response as uResponse, ResponseInit as uResponseInit } from 'undici';
+import { isDefined } from '../utils/nonNull';
+
+export class HttpResponse implements types.HttpResponse {
+    readonly cookies: types.Cookie[];
+    readonly enableContentNegotiation: boolean;
+
+    #uRes: uResponse;
+
+    constructor(resInit?: HttpResponseInit) {
+        const uResInit: uResponseInit = { status: resInit?.status, headers: resInit?.headers };
+        if (isDefined(resInit?.jsonBody)) {
+            this.#uRes = uResponse.json(resInit?.jsonBody, uResInit);
+        } else {
+            this.#uRes = new uResponse(resInit?.body, uResInit);
+        }
+
+        this.cookies = resInit?.cookies || [];
+        this.enableContentNegotiation = !!resInit?.enableContentNegotiation;
+    }
+
+    get status(): number {
+        return this.#uRes.status;
+    }
+
+    get headers(): Headers {
+        return this.#uRes.headers;
+    }
+
+    get body(): ReadableStream<any> | null {
+        return this.#uRes.body;
+    }
+
+    get bodyUsed(): boolean {
+        return this.#uRes.bodyUsed;
+    }
+
+    async arrayBuffer(): Promise<ArrayBuffer> {
+        return await this.#uRes.arrayBuffer();
+    }
+
+    async blob(): Promise<Blob> {
+        return await this.#uRes.blob();
+    }
+
+    async formData(): Promise<FormData> {
+        return await this.#uRes.formData();
+    }
+
+    async json(): Promise<unknown> {
+        return await this.#uRes.json();
+    }
+
+    async text(): Promise<string> {
+        return await this.#uRes.text();
+    }
+}
--- a/types/http.d.ts
+++ b/types/http.d.ts
@@ -3,12 +3,15 @@
 
 import { Blob } from 'buffer';
 import { ReadableStream } from 'stream/web';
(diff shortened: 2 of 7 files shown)


ejizba:
cc @aaronpowell

src/http/HttpResponse.ts
```diff
@@ -0,0 +1,64 @@
...
+        return await this.#uRes.formData();
+    }
+
+    async json(): Promise<unknown> {
+        return await this.#uRes.json();
+    }
+
+    async text(): Promise<string> {
+        return await this.#uRes.text();
+    }
```
aaronpowell, line 63: Do these need to do the async/await themselves, or could you just return the Promise, like so:

```suggestion
    blob(): Promise<Blob> {
        return this.#uRes.blob();
    }

    formData(): Promise<FormData> {
        return this.#uRes.formData();
    }

    json(): Promise<unknown> {
        return this.#uRes.json();
    }

    text(): Promise<string> {
        return this.#uRes.text();
    }

```

types/http.d.ts
```diff
@@ -191,13 +194,17 @@ export type HttpMethod = 'GET' | 'POST' | 'DELETE' | 'HEAD' | 'PATCH' | 'PUT' |
...
      * HTTP response body
      */
-    body?: HttpResponseBody;
+    body?: BodyInit;
+
+    /**
+     * A JSON-serializable HTTP Response body.
+     * If set, the `HttpResponseInit.body` property will be ignored in favor of this property
+     */
+    jsonBody?: any;
```
aaronpowell, line 207: Rather than having an `any`, why not make it a generic, so it's not straight returned it's assigned to a variable, you can get type information

ejizba: Do you have example code of how a generic would help here? I don't like generics 😝. If they can prevent a bug at compile time that normally would only be found at runtime, then they can be worth it. But most of the time I think they just give people a false sense of security and add unnecessary fluff to the codebase

src/http/HttpResponse.ts
ejizba: My concern was forgetting to await a return value in a `try/catch` which is why I made it a habit to add `await` everywhere. But I found a [lint rule](https://typescript-eslint.io/rules/return-await/) that can handle that situation. I'll turn on the lint rule in a separate PR as it might affect a lot of files

types/http.d.ts
aaronpowell: I was thinking of if you're writing a test and you want to assert on the output, you only have an `any` so you can't easily check the fields of the `jsonBody` for asserts.

hossam-nasr (approved):

Issue #12, opened 2022-09-02
Use an `app` root object for all function registrations


ejizba:
_Copied from https://github.com/ejizba/func-nodejs-prototype/issues/3_

Collecting feedback from various places on the `app` root object. Seems like everyone is overwhelmingly in favor of it, but we can have further discussion here.

From @aaronpowell:

> - I like this approach as it mirrors the kind of thing we find with Express, Fastify, and other web frameworks
> - It also makes for a clear ordering to the Functions that are defined, so if you've got multiple HTTP triggers that work off _similar_ routes, you should have a clear order of precedence
> - It would also make it easier to have discovery, rather than having to know what you can import, you import one thing and intellisense will tell you want it can do

From @YunchuWang

> With a central object like app for managing functions, it is more intuitive for users to know/look up methods compared to option1 via intellijsense. (ex, typing app. autocompletes list of methods)

ejizba:
More feedback from offline:

> - I think that exposing the framework through the app object might conflict with the user's code in some cases. The app variable name is common is user land.
> - I'd love to see the library exposed via an explicit name (and not requiring developers to unnecessarily rename it: const { app: funcApp } = require(...))

ejizba:
We modeled `app` after express, although the name `app` isn't as strict in their examples.
```javascript
const express = require('express')
const app = express() // technically I could name this anything
```

In regards to the second bullet point, users can already do this, which may be easier than the above workaround:
```javascript
const func = require(‘@azure/functions’);
func.app.
```

aaronpowell:
I do question this statement a bit:

> I think that exposing the framework through the app object might conflict with the user's code in some cases. The app variable name is common is user land.

Yes, it's true that `app` is a common when working with express/fastify/etc., but is it realistic that we'd expect the user to be having a codebase that blends both Functions and one of these other frameworks? I'd more be thinking that we'd have a porting story, so if you had an app like so:

```js
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```

We could have a porting story that allows you to write:

```js
const app = require('@azure/functions-express')
const port = 3000

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```

[...]

PR #50, opened 2023-01-06
Refactor namespaces to modules

ejizba:
Tangentially related to https://github.com/Azure/azure-functions-nodejs-library/issues/12, a lot of places (example [here](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html) and [here](https://github.com/microsoft/TypeScript/issues/30994#issuecomment-492024848)) say not to use namespaces for organizational reasons. Instead, they say use modules and re-export where necessary. In other words, instead of something like this:
```typescript
export namespace app { ... }
```

Just put the code in a separate file and re-export it like this:
```typescript
export * as app from './app';
```

At least, that's my interpretation. Functionally, this makes very little difference to our package (no changes necessary to the prototype). The PR is mostly just splitting code into separate files.


ejizba:
Fwiw, I did split this up into separate commits to make it easier to review. The first commit just moves the code. The second commit is what actually converted to namespaces and is the interesting piece to review. It should also be viewed with whitespace changes off: https://github.com/Azure/azure-functions-nodejs-library/pull/50/commits/920bbab2cca5a751e499995ce6d7fa5dc5fc76a9?w=1

hossam-nasr (approved):
LGTM! I also like that this reduces the length of our `index.ts` and `index.d.ts` and nicely separates code :)

Issue #52, opened 2023-01-25
Add method to list all bindings in v4


hossam-nasr:
In the old model, it was possible to list all the available bindings on the `context` object, by simply calling `context.bindings`.

In v4, this is how extra input/output bindings are retrieved:

```TS
const inputBinding = input.generic(/* whatever config */)
//...
// inside function definition
const inputValue = context.extraInputs.get(inputBinding);
```

Unlike before, there isn't anymore a way for the function to list all the available bindings. It's possible it may have other use-cases for customers too, but in Durable Functions, this allowed the Durable SDK to find the Durable Client input binding, by simply passing the invocation context, as below:

```TS
const client = context.df.getClient(context);
```

But now, the user also has to provide the input binding info, so the SDK is able to retrieve it:

```TS
const client = context.df.getClient(context, clientInput);
```

To make an experience comparable to the old model (see example here: https://github.com/Azure/azure-functions-durable-js/issues/414#issuecomment-1387674838), we would need a way for the SDK to retrieve all input bindings registered to the function, for example:

```TS
const allInputBindings = context.extraInputs.list();
```

PR #54, opened 2023-01-26
Add options to invocation context

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/52

The main goal is just to give people an easy way to inspect the configuration they used to register the function at the time of invocation.

### DIFF
--- a/src/InvocationModel.ts
+++ b/src/InvocationModel.ts
@@ -14,6 +14,7 @@ import {
 } from '@azure/functions-core';
 import { format } from 'util';
 import { returnBindingKey } from './constants';
+import { fromRpcBindings } from './converters/fromRpcBindings';
 import { fromRpcRetryContext, fromRpcTraceContext } from './converters/fromRpcContext';
 import { fromRpcTriggerMetadata } from './converters/fromRpcTriggerMetadata';
 import { fromRpcTypedData } from './converters/fromRpcTypedData';
@@ -53,6 +54,7 @@ export class InvocationModel implements coreTypes.InvocationModel {
             retryContext: fromRpcRetryContext(req.retryContext),
             traceContext: fromRpcTraceContext(req.traceContext),
             triggerMetadata: fromRpcTriggerMetadata(req.triggerMetadata, this.#triggerType),
+            options: fromRpcBindings(this.#bindings),
         });
 
         const inputs: unknown[] = [];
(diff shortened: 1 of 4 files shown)


types/InvocationContext.d.ts
```diff
@@ -90,6 +90,12 @@ export declare class InvocationContext {
      * For example, this will be undefined for http and timer triggers because you can find that information on the request & timer object instead
      */
     triggerMetadata?: TriggerMetadata;
+
+    /**
+     * The options used when registering the function
+     * NOTE: This value may differ slightly from the original because it has been validated and defaults may have been explicitly added
+     */
+    options: EffectiveFunctionOptions;
```
hossam-nasr, line 98: `EffectiveFunctionOptions` has `extraInputs` and `extraOuputs` as optional, when in reality, they're always there, even if initialized to empty arrays (as I think they should). Can we have another type for this that expresses that?

src/InvocationModel.ts
```diff
@@ -53,6 +54,7 @@ export class InvocationModel implements coreTypes.InvocationModel {
             retryContext: fromRpcRetryContext(req.retryContext),
             traceContext: fromRpcTraceContext(req.traceContext),
             triggerMetadata: fromRpcTriggerMetadata(req.triggerMetadata, this.#triggerType),
+            options: fromRpcBindings(this.#bindings),
```
hossam-nasr, line 57: My biggest issue with this implementation is that this now requires the `options` as input to the constuctor of `InvocationContext`, making the unit testing scenario harder. This code you have in the Wiki [here](https://github.com/Azure/azure-functions-nodejs-library/wiki/Azure-Functions-Node.js-Framework-v4) would no longer work:

```TS
const testInvocationContext = new InvocationContext({
  functionName: 'testFunctionName',
  invocationId: 'testInvocationId',
  logHandler: (_level, ...args) => console.log(...args)
});
```

Is there a way around this? Could we have the options object initialized inside the constructor, or set here after the `context = new InvocationContext({})` line?

types/InvocationContext.d.ts
ejizba: Done

src/InvocationModel.ts
ejizba: I assume this feedback applies to all properties on `InvocationContextInit`? I've updated so in testing scenarios you can just do `new InvocationContext()`

hossam-nasr: That's great! Yeah, I was gonna suggest that too but in a separate PR 😛

hossam-nasr (approved):
LGTM! Can we get a release of the library with these change so I can ingest it and make the necessary change on the SDK side?

Issue #53, opened 2023-01-25
Better handle return value when not set by a binding


ejizba:
Revisiting the code added in this PR: https://github.com/Azure/azure-functions-nodejs-library/pull/25

It "worked" in the old model for durable, but it's broken in the new model specifically when you have an additional extra output. It should still return the value, but it's not

PR #60, opened 2023-02-09
Fix bug for durable when returning value without a binding set

ejizba:
The old logic was based on the value itself and other bindings. The new logic is just "Did we use it yet?"

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/53


hossam-nasr (approved):
LGTM!

Issue #73, opened 2023-03-31
Support configuring retry policy when registering a function in v4


ejizba:
[Example `function.json` from v3](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-error-pages?tabs=fixed-delay%2Cin-process&pivots=programming-language-javascript#retry-examples)
```
{
    "bindings": [
        {
            ....
        }
    ],
    "retry": {
        "strategy": "fixedDelay",
        "maxRetryCount": 4,
        "delayInterval": "00:00:10"
    }
}
```

We don't have an analogous place for people to put the "retry" properties in v4.

PR #106, opened 2023-06-21
Add v4 support for specifying retry policy

ejizba:
This doc was particularly helpful as I was working on this feature: https://learn.microsoft.com/azure/azure-functions/functions-bindings-error-pages

Depends on https://github.com/Azure/azure-functions-nodejs-worker/pull/696

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/73

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,7 +6,9 @@ import {
     CosmosDBTrigger,
     EventGridFunctionOptions,
     EventHubFunctionOptions,
-    FunctionOptions,
+    ExponentialBackoffRetryOptions,
+    FixedDelayRetryOptions,
+    GenericFunctionOptions,
     HttpFunctionOptions,
     HttpHandler,
     HttpMethod,
@@ -21,6 +23,7 @@ import * as coreTypes from '@azure/functions-core';
 import { CoreInvocationContext, FunctionCallback } from '@azure/functions-core';
 import { InvocationModel } from './InvocationModel';
 import { returnBindingKey, version } from './constants';
+import { toRpcTimestamp } from './converters/toRpcNullable';
 import * as output from './output';
 import * as trigger from './trigger';
 import { isTrigger } from './utils/isTrigger';
@@ -231,7 +234,7 @@ export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void {
     });
 }
 
-export function generic(name: string, options: FunctionOptions): void {
+export function generic(name: string, options: GenericFunctionOptions): void {
     if (!hasSetup) {
         setup();
     }
@@ -271,12 +274,29 @@ export function generic(name: string, options: FunctionOptions): void {
         }
     }
 
+    let retryOptions: coreTypes.RpcRetryOptions | undefined;
+    if (options.retry) {
+        retryOptions = {
+            ...options.retry,
+            retryStrategy: options.retry.strategy,
+            delayInterval: toRpcTimestamp((<FixedDelayRetryOptions>options.retry).delayInterval, 'retry.delayInterval'),
+            maximumInterval: toRpcTimestamp(
+                (<ExponentialBackoffRetryOptions>options.retry).maximumInterval,
+                'retry.maximumInterval'
+            ),
+            minimumInterval: toRpcTimestamp(
+                (<ExponentialBackoffRetryOptions>options.retry).minimumInterval,
+                'retry.minimumInterval'
+            ),
+        };
+    }
+
     const coreApi = tryGetCoreApiLazy();
     if (!coreApi) {
         console.warn(
             `WARNING: Skipping call to register function "${name}" because the "@azure/functions" package is in test mode.`
         );
     } else {
-        coreApi.registerFunction({ name, bindings }, <FunctionCallback>options.handler);
+        coreApi.registerFunction({ name, bindings, retryOptions }, <FunctionCallback>options.handler);
     }
 }
--- a/src/app.ts
+++ b/src/app.ts
@@ -23,7 +23,7 @@ import * as coreTypes from '@azure/functions-core';
 import { CoreInvocationContext, FunctionCallback } from '@azure/functions-core';
 import { InvocationModel } from './InvocationModel';
 import { returnBindingKey, version } from './constants';
-import { toRpcTimestamp } from './converters/toRpcNullable';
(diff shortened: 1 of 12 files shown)


gunzip:
Since the retry policies will work only with kafka, timer and eventhub triggers I wonder if there's a way to extend only these specific options (shouldn't we?) in order to avoid to let users specify retry options for triggers where they won't work.

ejizba:
Hi @gunzip, I've already done that in this PR. I also have it as an option for Cosmos DB, per the docs I mentioned above. That being said, we don't want the Node.js model to be in charge of blocking people for things like this (I'm not sure when the other triggers will stop working), so I also added it to the `generic` options so that people can workaround us if they want to.

ejizba:
ping @hossam-nasr

types/generic.d.ts
```diff
@@ -1,6 +1,16 @@
...
+import { FunctionOptions, RetryOptions } from './index';
+
+export interface GenericFunctionOptions extends FunctionOptions {
+    /**
+     * An optional retry policy to rerun a failed execution until either successful completion occurs or the maximum number of retries is reached.
+     * Learn more [here](https://learn.microsoft.com/azure/azure-functions/functions-bindings-error-pages)
+     */
+    retry?: RetryOptions;
+}
+
```
hossam-nasr, line 13: Is there a reason this was not just added to all function options?

types/index.d.ts
```diff
@@ -112,3 +112,49 @@ export interface FunctionOutput extends Record<string, unknown> {
...
+    /**
+     * The maximum number of retries allowed per function execution. -1 means to retry indefinitely.
+     */
+    maxRetryCount: number;
+
+    /**
+     * The delay that's used between retries.
+     * This can be a Node.js Date or Unix time in milliseconds.
+     */
+    delayInterval: Date | number;
```
hossam-nasr: How would specifying a `Date` object work, if it's an interval not a specific date/time? Also, I think there is some value in how convenient it is to write intervals like `"HH:mm:ss"`

src/app.ts
```diff
@@ -271,12 +274,29 @@ export function generic(name: string, options: FunctionOptions): void {
         }
     }
 
+    let retryOptions: coreTypes.RpcRetryOptions | undefined;
+    if (options.retry) {
+        retryOptions = {
+            ...options.retry,
```
hossam-nasr, line 280: Is there a reason for this expansion?

types/generic.d.ts
ejizba: It's not supported in all trigger types. You can see the list of supported types here: https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-error-pages#retries. Specifically this stuff only applies when the "Retry Source" is "Retry policies" in the table shown at that link

types/index.d.ts
ejizba: That's a good point. I was kind of copying existing support we had for the "expires" property on http cookies, but that's an actual timestamp not a duration. I'll have to look into this

src/app.ts
ejizba: It's to copy all properties that don't need to be converted before sending over rpc. This is a common pattern I use, although retry options is somewhat unique in the sense that most properties need to be converted. I think `maxRetryCount` is the only one that doesn't

types/index.d.ts
ejizba: Updated to use a new duration interface instead of `Date`. I considered supporting the string "HH:mm:ss", but felt like an object was simpler for me to implement (no string parsing) and probably more type-safe/readable for users

src/converters/toRpcDuration.ts
```diff
@@ -0,0 +1,36 @@
...
+                const minutes = (dateTime.minutes || 0) + (dateTime.hours || 0) * 60;
+                const seconds = (dateTime.seconds || 0) + minutes * 60;
+                timeInMilliseconds = (dateTime.milliseconds || 0) + seconds * 1000;
+            } else if (typeof dateTime === 'number') {
+                timeInMilliseconds = dateTime;
+            }
+
+            if (isDefined(timeInMilliseconds) && timeInMilliseconds >= 0) {
+                return {
+                    seconds: Math.round(timeInMilliseconds / 1000),
```
hossam-nasr, line 23: Can't we use the `nanos` instead of `seconds` here, in order to not lose granularity?

hossam-nasr: What happens if I specify a value of less than 1 second?

types/index.d.ts
hossam-nasr: Yeah that's fair, although I would still argue it's better to be consistent with our docs, and perhaps there's a package that can already do the parsing for us? Regardless, I won't block on this

src/converters/toRpcDuration.ts
ejizba: The old model only supported the granularity up to second, so I think it's fine to do the same here. If the value is less than 1 it'll either round up or down to 1

If we want to improve the granularity I'd prefer to do it separately and I'd prefer to investigate adding "milliseconds" to the rpc types. Milliseconds is the de-facto standard that I'm used to, not nanos

types/index.d.ts
ejizba: We have to update the docs regardless for the v4 model samples, so shouldn't be a concern.

I prefer the interface over the string anyways so not sure how relevant it is, but yes there are packages to parse date/time strings (dayjs/luxon/moment). They have their own problems and it's not always worth adding packages even if we preferred that approach, though.

src/converters/toRpcDuration.ts
hossam-nasr: That's fair, but if we're only supporting granularity of 1 second, why are we allowing the customer to set the duration value in milliseconds not seconds? Since here we're allowing customers to specify a duration in milliseconds, it might hint that users can specify 400 ms for example. I'd prefer if either we let customers specify the numbre of seconds not milliseconds, or we throw an error or log a warning if a value would be rounded down to zero, for example if specifying 400 ms. What would even happen in that case?

hossam-nasr (approved):

src/converters/toRpcDuration.ts
ejizba: Discussed offline - should be good enough for GA as-is. We can always improve in the future

PR #115, opened 2023-07-20
Add basic hooks support (v4)

hossam-nasr:
This PR lays the groundwork to add basic support for hooks in v4. This is mostly a thin wrapper over the existing hooks Core API.

## Hooks supported

1. App Start hooks
2. App terminate hooks
3. Pre-Invocation hooks
4. Post-Invocation hooks

## Registering hooks

You can register the hook you desire as below:

```TS
const onStartHandler: AppStartHandler = async (_context: AppStartContext) => {
    console.log('this hook will be run at the start of the app.');
}
app.onStart(onStartHandler);

const onTerminateHandler: AppTerminateHandler = async (_context: AppTerminateContext) => {
    console.log('This hook will be run before grateful shutdown of the app.');
};
app.onTerminate(onTerminateHandler);

const preInvocHandler: PreInvocationHandler = async (_context: PreInvocationContext) => {
    console.log("This hook will be run before every invocation.");
}
app.onPreInvocation(preInvocHandler);

const postInvocHandler: PostInvocationHandler = async (_context: PostInvocationContext) => {
    console.log("This hook will be run after every invocation.");
}
app.onPostInvocation(postInvocHandler);
```

## Sharing data between hooks

The `hookData` and `appHookData` from the core API are replaced with `context.get`, `context.set` (replacing `hookData`), and `context.getGlobal` and `context.setGlobal` (replacing `appHookData`):

```TS
const onStartFunc: AppStartHandler = async (context: AppStartContext) => {
    // this will be available to all hooks
    context.setGlobal('globalKey', 'globalValue');
    context.set('localKey', 'localValue');
}
app.onStart(onStartFunc);

const preInvocHook: PreInvocationHandler = async (context: PreInvocationContext) => {
    assert(context.getGlobal('globalKey') === 'globalValue');
    assert(typeof context.get('appOnlyKey') === 'undefined');

    context.set('invocationOnlyKey', 'invocationOnlyValue');

}
app.onPreInvocation(preInvocHook);

const postInvocHook: PostInvocationHandler = async (context: PostInvocationContext) => {
    assert(context.getGlobal('globalKey') === 'globalValue');
    assert(typeof context.get('appOnlyKey') === 'undefined');
    assert(context.get('invocationOnlyKey') === 'invocationOnlyValue');
}
app.onPostInvocation(postInvocHook);

const onTerminateHook: AppTerminateHandler = async (context: AppTerminateContext) => {
    assert(context.get('appOnlyKey') === 'appOnlyValue');
    assert(context.getGlobal('globalKey') === 'globalValue');
    assert(typeof context.get('invocationOnlyKey') === 'undefined');
};
app.onTerminate(onTerminateHook);
```

### DIFF
--- a/types/app.d.ts
+++ b/types/app.d.ts
@@ -4,6 +4,7 @@
 import { CosmosDBFunctionOptions } from './cosmosDB';
 import { EventGridFunctionOptions } from './eventGrid';
 import { EventHubFunctionOptions } from './eventHub';
+import { AppStartHandler, AppTerminateHandler, Disposable, PostInvocationHandler, PreInvocationHandler } from './hooks';
 import { HttpFunctionOptions, HttpHandler, HttpMethodFunctionOptions } from './http';
 import { FunctionOptions } from './index';
 import { ServiceBusQueueFunctionOptions, ServiceBusTopicFunctionOptions } from './serviceBus';
@@ -150,3 +151,39 @@ export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
  * @param options Configuration options describing the inputs, outputs, and handler for this function
  */
 export function generic(name: string, options: FunctionOptions): void;
+
+/**
+ * Register a hook on the `appStart` event, executed at the start of your application
+ *
+ * @param handler the handler for the event
+ * @returns a `Disposable` object that can be used to unregister the hook
+ */
+export function onStart(handler: AppStartHandler): Disposable;
+
+/**
+ * Register a hook on the `appTerminate` event, executed during graceful shutdown of your application
+ * This hook will not be executed if your application is terminated forcefully
+ * Please note that all `appTerminate` hooks must finish execution in 10 seconds or less, or they will be terminated.
+ 
+ * @param handler the handler for the event
+ * @returns a `Disposable` object that can be used to unregister the hook
+ */
+export function onTerminate(handler: AppTerminateHandler): Disposable;
+
+/**
+ * Register a hook to be run right _before_ a function is invoked.
+ * This hook will be executed for all functions in your app.
+ *
+ * @param handler the handler for the event
+ * @returns a `Disposable` object that can be used to unregister the hook
+ */
+export function onPreInvocation(handler: PreInvocationHandler): Disposable;
+
+/**
+ * Register a hook to be run right _after_ a function is invoked.
+ * This hook will be executed for all functions in your app.
+ *
+ * @param handler the handler for the event
+ * @returns a `Disposable` object that can be used to unregister the hook
+ */
+export function onPostInvocation(handler: PostInvocationHandler): Disposable;
--- a/types/hooks.d.ts
+++ b/types/hooks.d.ts
@@ -1,0 +1,328 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { FunctionHandler } from '.';
+import { InvocationContext } from './InvocationContext';
+
+/**
+ * Represents a type which can release resources, such as event listening or a timer.
+ */
+export declare class Disposable {
+    /**
+     * Combine many disposable-likes into one. You can use this method when having objects with a dispose function which aren't instances of `Disposable`.
+     *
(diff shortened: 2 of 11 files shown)


types/hooks.d.ts
```diff
@@ -0,0 +1,328 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { FunctionHandler } from '.';
+import { InvocationContext } from './InvocationContext';
+
+/**
+ * Represents a type which can release resources, such as event listening or a timer.
+ */
+export declare class Disposable {
```
ejizba, line 10: `Disposable` is not intended to be unique to hooks. You could give it its own file, but since it's a simple class I'd probably just move it directly into `index.d.ts`

types/app.d.ts
```diff
@@ -150,3 +151,39 @@ export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
...
  */
 export function generic(name: string, options: FunctionOptions): void;
+
+/**
+ * Register a hook on the `appStart` event, executed at the start of your application
+ *
+ * @param handler the handler for the event
+ * @returns a `Disposable` object that can be used to unregister the hook
+ */
+export function onStart(handler: AppStartHandler): Disposable;
```
ejizba, line 161: I'm leaning towards this organization:
```
app.hook.start
app.hook.terminate
app.hook.preInvocation
app.hook.postInvocation
```

I'm not a big of "on". I feel like we could've easily done that for the trigger registrations (`app.onHttp`, `app.onTimer`), and I don't want to be inconsistent and start using "on" for hooks. That being said, I still like grouping all the hooks together to make the Intellisense cleaner. Rather than group them with the "on" prefix, I think a more explicit "hook" group would make sense.

It's a bit weird to have `hook` on `app` when `input`/`output`/`trigger` aren't on `app`, but I feel like it's okay because the latter don't actually register anything.

types/hooks.d.ts
```diff
@@ -0,0 +1,328 @@
...
+     */
+    get(propertyName: string): unknown;
+
+    /**
+     * Set a `propertyName` to some `value` to be shared across scopes for all hooks
+     *
+     * @param propertyName The name of the property to set
+     * @param value The value to set
+     */
+    setGlobal(propertyName: string, value: unknown): void;
```
ejizba, line 66: So I don't want us to feel obligated to follow the core api, especially since we know we want to clean up the hook data. Taking a step back, how is `setGlobal` and `getGlobal` better than just using a global variable?

Aka rather than this:
```typescript
const onStartFunc: AppStartHandler = async (context: AppStartContext) => {
    context.setGlobal('globalKey', 'test123');
}
app.onStart(onStartFunc);

const onTerminateHook: AppTerminateHandler = async (context: AppTerminateContext) => {
    assert(context.getGlobal('globalKey') === 'test123');
};
app.onTerminate(onTerminateHook);
```

users could do this:
```typescript
let globalValue: string;

const onStartFunc: AppStartHandler = async (context: AppStartContext) => {
    globalValue = 'test123';
}
app.onStart(onStartFunc);

const onTerminateHook: AppTerminateHandler = async (context: AppTerminateContext) => {
    assert(globalValue === 'test123');
};
app.onTerminate(onTerminateHook);
```

If the user is defining their own global variables, they can more easily set the type (aka `string` instead of `unknown`), and they don't have to deal with the key (mispelling, conflicting with another key, etc.).

Lastly, I think it would simplify our lives if we only need to worry about invocation-scoped hook data, which IMO is something users can't accomplish on their own so it's much more valuable.

types/hooks.d.ts
```diff
@@ -0,0 +1,328 @@
...
+     * The function handler for this specific invocation. Changes to this value _will_ affect the function itself
+     */
+    functionHandler: FunctionHandler;
+}
+
+/**
+ * Object passed to PreInvocationContext constructors.
+ * For testing purposes only
+ */
+export interface PreInvocationContextInit extends InvocationHookContextInit {
```
ejizba, line 213: I feel like you told me this already, but I'm confused why we have both `PreInvocationContextInit` and `PreInvocationCoreContext`, with "core" nested into "init". Can we have just one object without duplication/nesting?

ejizba:
Closing due to lack of activity

PR #146, opened 2023-09-18
Add support for table input & output

ejizba:
No one asked for this one, but all the languages use it as an example in the http trigger docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python-v2%2Cin-process%2Cnodejs-v4%2Cfunctionsv2&pivots=programming-language-javascript#using-route-parameters), so that's what inspired me to add it. Plus it was pretty low-hanging fruit.

Technically the table binding supports both "Azure Cosmos DB for Table" and "Azure Table Storage", so that's why it's by itself instead of being grouped with storage or cosmos.

Lastly, table doesn't support triggers at all. Just input and output.

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -8,6 +8,8 @@ import {
     GenericInputOptions,
     StorageBlobInput,
     StorageBlobInputOptions,
+    TableInput,
+    TableInputOptions,
 } from '@azure/functions';
 import { addBindingName } from './addBindingName';
 
@@ -15,6 +17,13 @@ export function storageBlob(options: StorageBlobInputOptions): StorageBlobInput 
     return addInputBindingName({
         ...options,
         type: 'blob',
+    });
+}
+
+export function table(options: TableInputOptions): TableInput {
+    return addInputBindingName({
+        ...options,
+        type: 'table',
     });
 }
 
--- a/src/output.ts
+++ b/src/output.ts
@@ -20,6 +20,8 @@ import {
     StorageBlobOutputOptions,
     StorageQueueOutput,
     StorageQueueOutputOptions,
+    TableOutput,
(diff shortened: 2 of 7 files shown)


hossam-nasr (approved):

LuwkasLima:
@ejizba thank you very much for looking at that. Has this been released already? I'm using 
"@azure/functions": "^4.0.0-alpha.1" w/ Azure-functions-core-tools@4 for MacOS w/ extension bundle [4.*, 5.0.0]. Can't see that in the library:

image

ejizba:
@LuwkasLima Yes, this PR has been released. We actually just released the GA version v4.0.0 of `@azure/functions`, so I highly recommend updating to that version instead of any "alpha" version

LuwkasLima:
Thx, I did update to the latest package and it worked as expected.

If I understood well, it doesn't handle upsert or delete at the moment. Do we have a roadmap for that to come?

ejizba:
The most I know is what [the docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-table-output?tabs=in-process%2Cnodejs-v4%2Ctable-api&pivots=programming-language-typescript) say:

> This output binding only supports creating new entities in a table. If you need to update an existing entity from your function code, instead use an Azure Tables SDK directly.

Seems like using the SDK directly is the recommended solution, so my guess is this is not on a roadmap.

This would not be specific to Node.js, which is what I focus on and what this repo focuses on. For the table-specific GitHub repo, I believe this is where it is: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables

Issue #176, opened 2023-10-19
Support warmup trigger


ejizba:
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript

Today you can use a warmup trigger with [generics](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Clinux%2Cazure-cli&pivots=nodejs-model-v4), but would be nice if people could do it using `app.warmup`.

I'm honestly not that familiar with the warmup trigger - I'm mostly filing this issue because we do see it used by Node.js customers in our telemetry. Further investigation is necessary

Issue #174, opened 2023-10-18
Wait to start consuming requests


charlesgardyn:
Is there a way to make the function do some execution before start receiving requests, like connecting to a database?

ejizba:
Hi @charlesgardyn what plan are you using? Consumption? Premium? A [warmup trigger](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript) might work, although it's not supported on all plans. I've filed this issue to track that work: https://github.com/Azure/azure-functions-nodejs-library/issues/176

If you're on consumption (like most people), we'd have to come up with an alternative. You can definitely run code at app startup in various ways, but requests might be starting to queue up, so not sure how much that helps.

charlesgardyn:
I'm using Premium

castrodd:
@charlesgardyn Warmup triggers are supported on Premium. Have you been able to try them for your use case? Let us know if that addresses your needs and I will close this issue once we hear back from you.

charlesgardyn:
It does, thank you!

PR #180, opened 2023-10-31
Add native support for warmup trigger

castrodd:
This PR is a fix for issues #174 and #176.

Currently, users wanting to set up a [warmup trigger](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript) need to use [generics](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Clinux%2Cazure-cli&pivots=nodejs-model-v4#generic-inputs-and-outputs). For example, they could write:

```
const { app, output, trigger } = require('@azure/functions');

app.generic('warmup', {
    trigger: trigger.generic({
        type: "warmupTrigger",
        direction: "in",
        name: "warmupContext"
    })
    handler: async (warmupContext, context) => {
        context.log('Function App instance is warm.');
    }
});

```

This works, but with the changes in this PR, users can now call the warmup trigger directly on `app`:

```
import { app, InvocationContext } from "@azure/functions";

export async function warmupFunction(warmupContext: object, context: InvocationContext): Promise<void> {
    context.log('Function App instance is warm.');
}

app.warmup('warmup', {
    handler: warmupFunction,
});

```

### DIFF
--- a/types/app.d.ts
+++ b/types/app.d.ts
@@ -9,6 +9,7 @@ import { HttpFunctionOptions, HttpHandler, HttpMethodFunctionOptions } from './h
 import { ServiceBusQueueFunctionOptions, ServiceBusTopicFunctionOptions } from './serviceBus';
 import { StorageBlobFunctionOptions, StorageQueueFunctionOptions } from './storage';
 import { TimerFunctionOptions } from './timer';
+import { WarmupFunctionOptions } from './warmup';
 
 /**
  * Registers an http function in your app that will be triggered by making a request to the function url
@@ -142,6 +143,13 @@ export function eventGrid(name: string, options: EventGridFunctionOptions): void
  * @param options Configuration options describing the inputs, outputs, and handler for this function
  */
 export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
+
+/**
+ * Registers a function in your app that will be triggered when an instance is added to scale a running function app
+ * @param name The name of the function. The name must be unique within your app and will mostly be used for your own tracking purposes
+ * @param options Configuration options describing the inputs, outputs, and handler for this function
+ */
+export function warmup(name: string, options: WarmupFunctionOptions): void;
 
 /**
  * Registers a generic function in your app that will be triggered based on the type specified in `options.trigger.type`
--- a/types/app.d.ts
+++ b/types/app.d.ts
@@ -145,7 +145,12 @@ export function eventGrid(name: string, options: EventGridFunctionOptions): void
 export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
 
 /**
- * Registers a function in your app that will be triggered when an instance is added to scale a running function app
+ * Registers a function in your app that will be triggered when an instance is added to scale a running function app.
+ * The warmup trigger is only called during scale-out operations, not during restarts or other non-scale startups.
+ * Make sure your logic can load all required dependencies without relying on the warmup trigger.
+ * Lazy loading is a good pattern to achieve this goal.
+ * The warmup trigger isn't available to apps running on the Consumption plan.
+ * For more information, please see the [Azure Functions warmup trigger documentation](https://learn.microsoft.com/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript).
  * @param name The name of the function. The name must be unique within your app and will mostly be used for your own tracking purposes
  * @param options Configuration options describing the inputs, outputs, and handler for this function
  */
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -68,7 +68,7 @@ export function eventGrid(options: EventGridTriggerOptions): EventGridTrigger;
 export function cosmosDB(options: CosmosDBTriggerOptions): CosmosDBTrigger;
 
 /**
- * [Link to docs and examples](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript)
+ * [Link to docs and examples](https://learn.microsoft.com/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript)
  */
 export function warmup(options: WarmupTriggerOptions): WarmupTrigger;
 
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -20,6 +20,7 @@ import {
     StorageQueueTriggerOptions,
 } from './storage';
 import { TimerTrigger, TimerTriggerOptions } from './timer';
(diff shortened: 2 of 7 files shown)


charlesgardyn:
Could you tell me what happens when you try to deploy that using the Consumption plan? It'd be nice if there were a nice error message

types/trigger.d.ts
```diff
@@ -66,6 +67,11 @@ export function eventGrid(options: EventGridTriggerOptions): EventGridTrigger;
  */
 export function cosmosDB(options: CosmosDBTriggerOptions): CosmosDBTrigger;
 
+/**
+ * [Link to docs and examples](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-warmup?tabs=isolated-process&pivots=programming-language-javascript)
```
ejizba: nit: Please remove `en-us/` whenever you add a link so that it defaults to the user's browser language instead of always english

types/warmup.d.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { FunctionOptions, FunctionResult, FunctionTrigger, RetryOptions } from './index';
+import { InvocationContext } from './InvocationContext';
+
+export type WarmupHandler = (warmupContext: object, context: InvocationContext) => FunctionResult;
```
ejizba: What does `warmupContext` actually look like when you run the function?

types/warmup.d.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface WarmupFunctionOptions extends WarmupTriggerOptions, Partial<FunctionOptions> {
+    handler: WarmupHandler;
+
+    trigger?: WarmupTrigger;
+
+    /**
+     * An optional retry policy to rerun a failed execution until either successful completion occurs or the maximum number of retries is reached.
+     * Learn more [here](https://learn.microsoft.com/azure/azure-functions/functions-bindings-error-pages)
+     */
+    retry?: RetryOptions;
```
ejizba: Retry isn't officially supported for all trigger types. I don't see warmup listed in the docs, so let's remove this property

https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-error-pages?tabs=fixed-delay%2Cisolated-process&pivots=programming-language-javascript

types/app.d.ts
```diff
@@ -143,6 +144,13 @@ export function eventGrid(name: string, options: EventGridFunctionOptions): void
  */
 export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
 
+/**
+ * Registers a function in your app that will be triggered when an instance is added to scale a running function app
+ * @param name The name of the function. The name must be unique within your app and will mostly be used for your own tracking purposes
```
ejizba, line 154: I think we should include more of the "considerations" listed in the warmup trigger docs. Maybe something like this:

> Registers a function in your app that will be triggered during scale-out operations.
> It's not triggered during restarts or other non-scale startups, so make sure your app can run without relying on the warmup trigger. Lazy loading is a good pattern to achieve this goal.
> The warmup trigger isn't available on the Consumption plan.

types/warmup.d.ts
castrodd: I am not able to glean much information from the runtime logs. All I can see is that it is an empty object with no prototype. I compared the types used by the other languages in the docs and they seem to be using a simple object as well for the type signature. I can't think of a way to gather more information at the moment. Do you have any tips?

ejizba: You may have done this already, but I would just do `context.log(JSON.stringify(warmupContext));`

types/trigger.d.ts
castrodd: Thanks for catching that. I fixed the url.

types/warmup.d.ts
castrodd: Yeah, I had removed it but then I flip-flopped. Removed it for good now.

types/app.d.ts
castrodd: Good point. I added more information and a link to the source as well.

types/warmup.d.ts
castrodd: Yeah, I tried, but all it logs out is an empty object: `{}`.

ejizba: Okay I feel like `object` makes it seem like we just forgot the types. I'd rather do one of the following:
1. Add a placeholder/empty interface like [this one](https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/types/eventGrid.d.ts#L18)
2. Somehow change the WarmupHandler so that it doesn't pass the input and only passes the InvocationContext

castrodd: I agree. Just having `object` does not look right. I would like to do option two, but I see a `warmupContext` in the docs for all the other languages, so it must be there for a reason (I hope). 

I added an empty interface for now. I can keep digging to see if I find more information on what the interface should actually look like.

types/warmup.d.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { FunctionOptions, FunctionResult, FunctionTrigger } from './index';
+import { InvocationContext } from './InvocationContext';
+
+export interface WarmupContextOptions {}
```
ejizba, line 7: nit: Should be just `WarmupContext` without `Options`

types/warmup.d.ts
ejizba: > so it must be there for a reason (I hope).

I kinda doubt it. Every other trigger type has an input - my guess is they just did an empty object because it was easier than breaking the pattern.

types/app.d.ts
```diff
@@ -143,6 +144,18 @@ export function eventGrid(name: string, options: EventGridFunctionOptions): void
  */
 export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void;
 
+/**
```
ejizba, line 147: > Could you tell me what happens when you try to deploy that using the Consumption plan? It'd be nice if there were a nice error message

@charlesgardyn this npm package doesn't have visibility into what plan its running under, so unfortunately I don't think it's possible for us to do anything. For the record it would be at most a warning, since your app needs to be able to run without the warmup trigger regardless of plan type. The [functions host](https://github.com/Azure/azure-functions-host) might be able to do the warning, but I don't see warmup triggers used often enough that I feel like bothering them

ejizba (approved):

types/app.d.ts
charlesgardyn: I see, I thought that was possible. It's ok if not, it's for some specific use cases indeed

PR #181, opened 2023-11-01
Add basic hook support

### LINKED ISSUE #7: Add hooks api to framework package
At the moment hooks are only exposed in the core api. However, we want the new programming model packages to wrap the core api and expose a "nicer" version of the hooks api (better intellisense, maybe better naming, full docs, etc.)

ejizba:
This PR supports app start, app terminate, pre invocation, and post invocation hooks. Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/7

To see example code, check out the end-to-end tests I added [here](https://github.com/Azure/azure-functions-nodejs-e2e-tests/pull/26) in `app/v4/src/index.ts`. I also added unit tests in this repo, but they're less meaningful (they mainly ensure hooks are unit-testable by our users as well).

Loosely based off Hossam's work in https://github.com/Azure/azure-functions-nodejs-library/pull/115, which has been delayed a lot. I think part of the delay was from feature-creep, so my focus is just to get the MVP out and I don't have any plans to add all the [extra features](https://github.com/Azure/azure-functions-nodejs-library/issues?q=is%3Aopen+is%3Aissue+label%3Ahooks) at this time. I'm super excited about hooks even in MVP-form and I think our users will be, too.

### DIFF
--- a/src/hooks/registerHook.ts
+++ b/src/hooks/registerHook.ts
@@ -1,0 +1,51 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AppStartHandler, AppTerminateHandler, PostInvocationHandler, PreInvocationHandler } from '@azure/functions';
+import * as coreTypes from '@azure/functions-core';
+import { Disposable } from '../utils/Disposable';
+import { tryGetCoreApiLazy } from '../utils/tryGetCoreApiLazy';
+import { AppStartContext } from './AppStartContext';
+import { AppTerminateContext } from './AppTerminateContext';
+import { PostInvocationContext } from './PostInvocationContext';
+import { PreInvocationContext } from './PreInvocationContext';
+
+function registerHook(hookName: string, callback: coreTypes.HookCallback): coreTypes.Disposable {
+    const coreApi = tryGetCoreApiLazy();
+    if (!coreApi) {
+        console.warn(
+            `WARNING: Skipping call to register ${hookName} hook because the "@azure/functions" package is in test mode.`
+        );
+        return new Disposable(() => {
+            console.warn(
+                `WARNING: Skipping call to dispose ${hookName} hook because the "@azure/functions" package is in test mode.`
+            );
+        });
+    } else {
+        return coreApi.registerHook(hookName, callback);
+    }
+}
+
+export function appStart(handler: AppStartHandler): Disposable {
+    return registerHook('appStart', (coreContext) => {
+        return handler(new AppStartContext(coreContext));
+    });
+}
+
+export function appTerminate(handler: AppTerminateHandler): Disposable {
+    return registerHook('appTerminate', (coreContext) => {
+        return handler(new AppTerminateContext(coreContext));
+    });
+}
+
+export function preInvocation(handler: PreInvocationHandler): Disposable {
+    return registerHook('preInvocation', (coreContext) => {
+        return handler(new PreInvocationContext(coreContext));
+    });
+}
+
+export function postInvocation(handler: PostInvocationHandler): Disposable {
+    return registerHook('postInvocation', (coreContext) => {
+        return handler(new PostInvocationContext(coreContext));
+    });
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,10 +1,17 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
+export { InvocationContext } from './InvocationContext';
 export * as app from './app';
+export { AppStartContext } from './hooks/AppStartContext';
+export { AppTerminateContext } from './hooks/AppTerminateContext';
+export { HookContext } from './hooks/HookContext';
+export { InvocationHookContext } from './hooks/InvocationHookContext';
+export { PostInvocationContext } from './hooks/PostInvocationContext';
+export { PreInvocationContext } from './hooks/PreInvocationContext';
 export { HttpRequest } from './http/HttpRequest';
 export { HttpResponse } from './http/HttpResponse';
 export * as input from './input';
-export { InvocationContext } from './InvocationContext';
 export * as output from './output';
 export * as trigger from './trigger';
+export { Disposable } from './utils/Disposable';
--- a/src/utils/Disposable.ts
+++ b/src/utils/Disposable.ts
@@ -2,8 +2,8 @@
 // Licensed under the MIT License.
 
 /**
- * Based off of the Node worker
- * https://github.com/Azure/azure-functions-nodejs-worker/blob/bf28d9c5ad4ed22c5e42c082471d16108abee140/src/Disposable.ts
+ * Based off of VS Code
+ * https://github.com/microsoft/vscode/blob/7bed4ce3e9f5059b5fc638c348f064edabcce5d2/src/vs/workbench/api/common/extHostTypes.ts#L65
  */
 export class Disposable {
     static from(...inDisposables: { dispose(): any }[]): Disposable {
@@ -27,7 +27,7 @@ export class Disposable {
     }
 
     dispose(): any {
-        if (this.#callOnDispose instanceof Function) {
+        if (typeof this.#callOnDispose === 'function') {
             this.#callOnDispose();
             this.#callOnDispose = undefined;
         }
--- a/src/utils/Disposable.ts
+++ b/src/utils/Disposable.ts
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Based off of the Node worker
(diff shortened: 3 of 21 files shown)


src/utils/Disposable.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Based off of the Node worker
+ * https://github.com/Azure/azure-functions-nodejs-worker/blob/bf28d9c5ad4ed22c5e42c082471d16108abee140/src/Disposable.ts
+ */
+export class Disposable {
+    static from(...inDisposables: { dispose(): any }[]): Disposable {
```
castrodd, line 9: I see in the link above that we have the same call signature, but I noticed in [this file](https://github.com/Azure/azure-functions-nodejs-library/blob/c3aa2c4c2cff28317e00ff23460758b47c7c47f5/types-core/index.d.ts#L159) that it is slightly different. Should we align all three? Or is there a reason the example in `types-core` is slightly different?

ejizba: It's actually based on VS Code, not from the worker. I'll update the comment and get the latest VS Code source. They have the same inconsistency (source [here](https://github.com/microsoft/vscode/blob/7bed4ce3e9f5059b5fc638c348f064edabcce5d2/src/vs/workbench/api/common/extHostTypes.ts#L65) and type [here](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L1675)). I don't think the inconsistency between the source and the type matters in a functional sense, so I'd rather stay consistent between us and VS Code

src/hooks/registerHook.ts
```diff
@@ -0,0 +1,51 @@
...
+import { AppStartContext } from './AppStartContext';
+import { AppTerminateContext } from './AppTerminateContext';
+import { PostInvocationContext } from './PostInvocationContext';
+import { PreInvocationContext } from './PreInvocationContext';
+
+function registerHook(hookName: string, callback: coreTypes.HookCallback): coreTypes.Disposable {
+    const coreApi = tryGetCoreApiLazy();
+    if (!coreApi) {
+        console.warn(
+            `WARNING: Skipping call to register ${hookName} hook because the "@azure/functions" package is in test mode.`
```
castrodd, line 17: Nit: should we extract into a simple function so we don't need to repeat the string and also in case we need to edit in the future? 

```
const warningMessage = (hookName: string) => `WARNING: Skipping call to register ${hookName} hook because the "@azure/functions" package is in test mode.`;
```

src/index.ts
```diff
@@ -1,10 +1,17 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
+export { InvocationContext } from './InvocationContext';
```
castrodd, line 4: Why did InvocationContext jump to the top of the list?

types/hooks/HookContext.d.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Base class for all hook context objects
+ */
+export declare class HookContext {
+    /**
+     * For testing purposes only. This will always be constructed for you when run in the context of the Azure Functions runtime
```
castrodd, line 9: Nit: missing period after `runtime`

types/hooks/HookContext.d.ts
```diff
@@ -0,0 +1,27 @@
...
+     */
+    constructor(init?: HookContextInit);
+
+    /**
+     * The recommended place to store and share data between hooks in the same scope (app-level vs invocation-level).
+     * You should use a relatively unique property name so that it doesn't conflict with other hooks' data.
```
castrodd: Nit: I think it's better to drop `relatively`. The more unique, the safer, right?

types/hooks/HookContext.d.ts
```diff
@@ -0,0 +1,27 @@
...
+     * This object is readonly. You may modify it, but attempting to overwrite it will throw an error
```
castrodd, line 16: Nit: missing period after `error`

types/hooks/appHooks.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { HookContext, HookContextInit } from './HookContext';
+
+/**
+ * Handler for app start hooks
+ */
+export type AppStartHandler = (context: AppStartContext) => void | Promise<void>;
+
+/**
+ * Context on a function app during app startup.
```
castrodd, line 12: There's inconsistency with the doc comments in this file. Sometimes they end with a period, sometimes they don't. We should try to remain consistent.

test/hooks.test.ts
```diff
@@ -0,0 +1,79 @@
...
+describe('hooks', () => {
+    it("register doesn't throw error in unit test mode", () => {
+        app.hook.appStart(() => {});
+        app.hook.appTerminate(() => {});
+        app.hook.postInvocation(() => {});
+        const registeredHook = app.hook.preInvocation(() => {});
+        registeredHook.dispose();
+    });
+
+    function validateHookContext(context: HookContext) {
```
castrodd: Having the helper functions in the middle of the `describe` block makes it a little harder to read through the tests. Maybe it might be easier to have them at the end of the block? We do something similar here: [test/errors.test.ts](https://github.com/Azure/azure-functions-nodejs-library/blob/f5af37479bb1327809756cf628456c33aed9e2ab/test/errors.test.ts)

castrodd (approved):

types/hooks/appHooks.d.ts
ejizba: Yeah I go back and forth in my own head on this. Overall I like using a period, but I also feel weird using a period when the comment isn't a full sentence

Regardless, for stylistic choices like this I usually say we enforce it automatically or we don't worry about it. We use eslint/prettier to do that, but unfortunately I'm not finding a rule to enforce a period (or no period) at the end of a doc comment

test/hooks.test.ts
ejizba: I don't think the rule is as simple as "helper functions at the end", but I see what you mean for this file. I like to have code close to where it's actually used so in some cases I like the helper functions in the middle.

src/hooks/registerHook.ts
ejizba: Well the messages are different - one says "register" and one says "dispose". On my old team we would get our strings translated into other languages, and the translators preferred a little duplication over logic that dynamically constructed messages because the dynamic logic didn't work in other languages. We don't translate our strings, but I think I'm fine keeping "a little duplication" as a habit - at minimum it makes it easier to "ctrl+f" a string in the repo

src/index.ts
ejizba: No idea. We've got a few settings to automatically organize these, so I normally just let it do what it wants

For example: https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/.vscode/settings.json#L4

types/hooks/appHooks.d.ts
castrodd: For clauses like `Context on a function app during app startup`, we don't need to worry about it. There are some full sentences with a period and some without. I think those should be the only ones we try to clean up. I don't think ESLint or Prettier will cover grammatical/style issues like this, however.

src/hooks/registerHook.ts
castrodd: Oops, I should have read more carefully. I missed the slight change in wording. Yeah, in this case, there is no real benefit then.

src/index.ts
castrodd: Hmm, I tried to understand the rule. It seems VS Code uses ESLint's rules: https://eslint.org/docs/latest/rules/sort-imports. But reading that page, I still don't see how this happened. 😕

types/hooks/appHooks.d.ts
ejizba: To clarify, I agree a period is technically correct, but it's not my default style choice (see my use of periods in this thread so far - I swear I didn't do that on purpose lol). I'm probably the source of the sentences without periods and Hossam probably did the other ones. Periods aren't required for understanding the content and imply a certain level of formality/seriousness/emotion that I prefer to avoid. I'm [not alone](https://lifehacker.com/dont-use-periods-in-texts-1843744818) in this habit, either

So if we find an automated eslint/prettier rule I'm willing to go against my style preference, but otherwise I don't want to change my habits for something I don't necessarily like anyways

src/index.ts
ejizba: I think it's because of the casing. "InvocationContext" is capital and I guess that means it prefers it before all lower case ones. Switching to use the eslint rule directly instead of through VS Code probably makes sense, but I'll look into this separate from this PR because I don't want to clutter the diff

types/hooks/appHooks.d.ts
ejizba: Merging as-is, but we can continue to discuss outside this PR if you want

PR #186, opened 2023-11-02
Create codeql.yml

castrodd:
This PR is enabling CodeQL scanning. More information can be found here: https://codeql.github.com/docs/.

I have kept all the defaults except for the `schedule`. I adjusted this to every Wednesday at 12:30 (UTC).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,82 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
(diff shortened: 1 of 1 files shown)


ejizba (approved):
Don't know what most of that means, but willing to try it 🕵🏻‍♂️

castrodd:
> Don't know what most of that means, but willing to try it 🕵🏻‍♂️

Let's see if we like it. If we don't, then deleting this file should turn off code scanning again.

PR #188, opened 2023-11-06
Switch to eslint-based import sorting

ejizba:
Related to [this discussion](https://github.com/Azure/azure-functions-nodejs-library/pull/181#discussion_r1380618426), I wanted to switch off of VS Code's organize imports feature and onto an eslint-based one. I split this PR into two commits for easier review: one to configure eslint and one to auto-fix the repo based on the new config.

## Background (feel free to ignore if not interested)

I believe VS Code is using a built-in feature of TypeScript's language service that sorts imports (not an eslint rule). I can't find the feature documented by TypeScript, but it was announced [here](https://devblogs.microsoft.com/typescript/announcing-typescript-2-8-2/#organize-imports).

Turns out there are a ton of ways to do this 😅. These are the options I considered:

1. eslint's built-in [sort-imports](https://eslint.org/docs/latest/rules/sort-imports) rule: very basic features with no auto-fix
1. [prettier-plugin-organize-imports](https://www.npmjs.com/package/prettier-plugin-organize-imports): uses the TypeScript language service and explicitly mentions it should be the same as VS Code's feature. Not a super popular package, though
1. [eslint-plugin-import](https://www.npmjs.com/package/eslint-plugin-import): Lots of different rules and options, some of which support auto-fix
1. [eslint-plugin-simple-import-sort](https://www.npmjs.com/package/eslint-plugin-simple-import-sort): A more opinionated package that prioritizes auto-fix

I liked option 4 the best. "auto-fix" is a priority for me on this and they knocked it out of the park. I followed their [example config](https://github.com/lydell/eslint-plugin-simple-import-sort/#example-configuration) and made one tweak to [remove whitespace between groups](https://github.com/lydell/eslint-plugin-simple-import-sort/#how-do-i-remove-all-blank-lines-between-imports).


castrodd (approved):
Looks great!

.eslintrc.json
```diff
@@ -3,7 +3,9 @@
     "plugins": [
         "@typescript-eslint",
         "header",
-        "deprecation"
+        "deprecation",
```
castrodd, line 6: Tell me if I am going overboard here, but should we consider sorting this list of plugins and also the parserOptions? The parserOptions list in particular is getting quite long and it might be good to bring some structure to it.

.eslintrc.json
```diff
@@ -3,7 +3,9 @@
...
+        "simple-import-sort",
+        "import"
```
castrodd, line 8: You mentioned in the Description that you liked `simple-import-sort` the best. Were we not able to get everything we need from that package? Is that why we also need this second `import` plugin?

ejizba: `simple-import-sort` recommended the second plugin in their [example config](https://github.com/lydell/eslint-plugin-simple-import-sort/#example-configuration), so I still count it as a part of option 4. And yeah it added some extra stuff. For example, the `import` plugin will combine these into one:

```
import { EventGridFunctionOptions } from '@azure/functions';
import { CosmosDBFunctionOptions } from '@azure/functions';
```

.eslintrc.json
ejizba: For some of these config options the order matters, although I can't remember which. For example if things conflict with each other (i.e. eslint and prettier might format things a little different) the one listed last is the one that takes effect

Btw, I think you mean the "rules" section, not the "parserOptions" section. The file is only 64 lines long, so I'm not worried about it 😆 You're welcome to organize it

castrodd: Yes, you're right. I meant the "rules" section. The diff made it seem like the "rules" were under "parserOptions". I can sort them once you merge this PR.

PR #198, opened 2023-11-16
Add Resource Management Policy

castrodd:
This PR is adding a Resource Management Policy. Previously, the work done by these policies was performed by Fabric Bot, but that service was deprecated earlier this year.

More information about Fabric Bot and its replacement can be found [here](https://docs.opensource.microsoft.com/tools/fabric-bot/) (this link will only be accessible to Microsoft employees).

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -11,40 +11,40 @@ configuration:
     - description: 
       frequencies:
       - hourly:
-          hour: 6
+          hour: 3
       filters:
       - isIssue
       - isOpen
       - hasLabel:
-          label: needs-author feedback
+          label: 'Needs: Author Feedback'
       - hasLabel:
           label: no recent activity
       - noActivitySince:
-          days: 7
+          days: 3
       actions:
       - closeIssue
     - description: 
       frequencies:
       - hourly:
-          hour: 6
+          hour: 3
       filters:
       - isIssue
       - isOpen
       - hasLabel:
-          label: needs-author feedback
+          label: 'Needs: Author Feedback'
       - noActivitySince:
-          days: 5
+          days: 4
       - isNotLabeledWith:
           label: no recent activity
       actions:
       - addLabel:
           label: no recent activity
       - addReply:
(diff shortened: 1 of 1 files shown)


.github/policies/resourceManagement.yml
```diff
@@ -0,0 +1,222 @@
+id:
```
ejizba, line 1: Which repo did you copy this file from?

castrodd: I took it from this PR: https://github.com/Azure/azure-functions-python-worker/pull/1285

castrodd: Based on the samples [here](https://eng.ms/docs/more/github-inside-microsoft/policies/resource-management), I don't think we need `id` or `owner`. Should I delete them?

ejizba: There's some inconsistencies in terms of the labels between the Python repo and what we use on Node.js repos. For example "needs-author feedback" vs "Needs: Author Feedback". I think the Node.js worker repo copied the labels from the host repo originally (example [here](https://github.com/Azure/azure-functions-host/blob/dev/.github/policies/resourceManagement.yml#L19)) and I stayed consistent with that when I created the library repo labels. I think the host repo is one of the more well-maintained repos for our team, so I kinda like staying consistent with them.

I'd start by copying the node worker [bot file](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/.github/policies/resourceManagement.yml) instead of the python one. Unfortunately the bot doesn't work in that repo (example issue [here](https://github.com/Azure/azure-functions-nodejs-worker/issues/727) which I think the bot should've closed). So we'd need to figure that out, too

castrodd: I updated the file to follow the Node worker repo. I also reached out to the Github in Microsoft team to see if they can help us understand why the policy is not working.

ejizba (approved):
I'm okay to merge, but please make sure to monitor issues closely and see if the bot is doing what we want (if it's doing anything at all)

castrodd:
Confirmed that the bot is working correctly on the Worker repo. I expect this repo will work the same way now. Merging PR.

PR #178, opened 2023-10-25
Introduce "connection" property

castrodd:
This PR is a fix for issue #177.

Currently, customers must use a topic key to authenticate to Event Grid when using an Event Grid output binding. These changes allow customers to also leverage the "connection" property to implement identity-based authentication.

For more details on these two different authentication methods, please refer to the following [documentation](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid-output?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cextensionv3&pivots=programming-language-typescript#connections).

PR #199, opened 2023-11-22
Add Either type for EventGrid output connections

castrodd:
In PR #178 we faced the problem that there are two exclusive ways to configure an EventGrid output binding: key-based or identity-based.

However, TypeScript does not have a built-in utility type that forces users to implement one interface or another but not both. We left the code in a state where users would be warned if they:

1. Used `topicEndpointUri` or `topicKeySetting` but not both
2. Did not specify any properties

But users were free to combine the `connection` property with `topicEndpointUri` or `topicKeySetting` or both without any warnings or errors thrown by the type checker. Our only defense against this was a comment that `When setting the connection property, the topicEndpointUri and topicKeySetting properties should NOT be set.`

Ideally, this comment would be unnecessary and our type design would prevent any such misconfigurations.

This PR introduces a custom utility type `Either`. It takes two interfaces A and B and allows two possible implementations: all the properties in A are accepted but none of the properties in B or vice-versa.

Now if a user tries to supply a `connection` property along with key-based settings, this usage will be flagged:

image
`Type '{ topicEndpointUri: string; topicKeySetting: string; connection: string; }' is not assignable to type '{ topicEndpointUri?: never; topicKeySetting?: never; }'.`

image
`Type '{ topicEndpointUri: string; connection: string; }' is not assignable to type '{ topicEndpointUri?: never; topicKeySetting?: never; }'.`

image
`Type '{ topicKeySetting: string; connection: string; }' is not assignable to type '{ topicEndpointUri?: never; topicKeySetting?: never; }'.`

And the correct implementations of both `EventGridOutputKeyOptions` and `EventGridOutputConnectionOptions` still work as expected:

image

image


types/utils/either.ts
```diff
@@ -0,0 +1,3 @@
+type Only<T, U> = { [P in keyof T]: T[P] } & { [P in keyof U]?: never };
```
ejizba, line 1: This breaks down if T and U share a key, causing the overall type to be `never`:

Screenshot 2023-11-28 at 10 46 15 AM

types/utils/either.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export type Either<T, U> = Only<T, U> | Only<U, T>;
```
ejizba, line 3: I like that there's an error, although I'm not in love with the message - it's a bit confusing. If you haven't already, can you check for:
1. Any open TypeScript issues where they discuss this. I'm curious if they have plans to change the behavior of the union type or why they haven't already
2. Any eslint rules a customer could use to cause the error themselves

Also, can you confirm that the messages for the existing errors remain the same and haven't regressed?

types/utils/either.ts
castrodd: I found a modification of `Only` that accounts for these cases:

```
type Only<T, U> = {   [P in keyof T]: T[P] } & Omit<{   [P in keyof U]?: never }, keyof T>
```

types/utils/either.ts
castrodd: Okay, I tried to find any issues that discuss this behavior. From everything I could find, this seems to be the intended behavior of union types. I would not expect any changes there.

ESLint does not have any particular rules to cover this, but typescript-eslint came close. There were 4 rules there involving union types:

1. Disallow duplicate constituents of union or intersection types
2. Disallow members of unions and intersections that do nothing or override type information
3. Enforce constituents of a type union/intersection to be sorted alphabetically
4. Disallow two overloads that could be unified into one with a union or an optional/rest parameter

None of those rules specifically address this scenario however.

I think you are right about the overall goal. Even though it is preferable to prevent users from doing the wrong thing, we also want them to understand why it is the wrong thing.

The error messages produced by these new types **are** very confusing. For example, the modified `Only` type gives me this error message when I use `topicEndpointUri` and `connection`:

```
Argument of type '{ topicEndpointUri: string; connection: string; }' is not assignable to parameter of type 'EventGridOutputOptions'.
  Type '{ topicEndpointUri: string; connection: string; }' is not assignable to type 'Only<EventGridOutputConnectionOptions, EventGridOutputKeyOptions>'.
    Type '{ topicEndpointUri: string; connection: string; }' is not assignable to type 'Omit<{ topicEndpointUri?: never; topicKeySetting?: never; }, "connection">'.
      Types of property 'topicEndpointUri' are incompatible.
        Type 'string' is not assignable to type 'never'.ts(2345)
```

That is truly awful. 

It's much clearer to keep the comment we have now that warns users. If they hit problems, I expect most will see the comment and that will take care of things.

I can abandon this PR and we can revisit this if it ever becomes a recurring issue for users.

PR #209, opened 2024-01-10
Add support for request & response clone

### LINKED ISSUE #207: How to reference HttpRequest body in hooks
@ejizba I understand the design choice and looking forward to stream support. However, it would be beneficial to have an option to cache the request body. I'm writing a hook that validates the request body. The hook consumes it using `.json()`. Now the function has no way to access the request body. Because everything is readonly in `HttpRequest` I also don't see an easy option to reset/replace the consumed stream. Can you provide guidance how a `preInvocation` hook and a function can use the request body?

*Update*: I've tried to use `pipeThrough`, but that still causes the `bodyUsed` flag to be true. Looks like the only option is to copy the entire request:

```ts
  const textBody = await request.text()
  parsedBody = JSON.parse(textBody)

  // TODO: validate body here

  const headerCopy: Record<string, string> = {}
  origRequest.headers?.forEach((value, key) => {
    headerCopy[key] = value
  })

  const queryParams: Record<string, string> = {}
  origRequest.query?.forEach((value, key) => {
    queryParams[key] = value
  })

  request = new HttpRequest({
    method: origRequest.method,
    url: origRequest.url,
    body: { string: textBody },
    headers: headerCopy,
    query: queryParams,
    params: origRequest.params,
  })
```

_Originally posted by @restfulhead in https://github.com/Azure/azure-functions-nodejs-library/issues/79#issuecomment-1875214147_

ejizba:
Just mirroring the existing "clone" methods already supported in undici/fetch. Per [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/clone), "the main reason clone() exists is to allow multiple uses of body objects (when they are one-use only.)"

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/207

### DIFF
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -27,11 +27,11 @@ export class HttpRequest implements types.HttpRequest {
     constructor(init: InternalHttpRequestInit) {
         this.#init = init;
 
-        const url = nonNullProp(init, 'url');
-
         if (init.undiciRequest) {
             this.#uReq = init.undiciRequest;
         } else {
+            const url = nonNullProp(init, 'url');
+
             let body: Buffer | string | undefined;
             if (init.body?.bytes) {
                 body = Buffer.from(init.body?.bytes);
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -12,31 +12,42 @@ import { fromNullableMapping } from '../converters/fromRpcNullable';
 import { nonNullProp } from '../utils/nonNull';
 import { extractHttpUserFromHeaders } from './extractHttpUserFromHeaders';
 
+interface InternalHttpRequestInit extends RpcHttpData {
+    undiciRequest?: uRequest;
+}
+
 export class HttpRequest implements types.HttpRequest {
     readonly query: URLSearchParams;
     readonly params: HttpRequestParams;
 
     #cachedUser?: HttpRequestUser | null;
     #uReq: uRequest;
-    #body?: Buffer | string;
+    #init: InternalHttpRequestInit;
 
-    constructor(rpcHttp: RpcHttpData) {
-        const url = nonNullProp(rpcHttp, 'url');
+    constructor(init: InternalHttpRequestInit) {
+        this.#init = init;
 
-        if (rpcHttp.body?.bytes) {
-            this.#body = Buffer.from(rpcHttp.body?.bytes);
-        } else if (rpcHttp.body?.string) {
-            this.#body = rpcHttp.body.string;
+        const url = nonNullProp(init, 'url');
+
+        if (init.undiciRequest) {
+            this.#uReq = init.undiciRequest;
+        } else {
+            let body: Buffer | string | undefined;
+            if (init.body?.bytes) {
+                body = Buffer.from(init.body?.bytes);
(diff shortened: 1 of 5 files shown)


src/http/HttpRequest.ts
```diff
@@ -86,4 +97,10 @@ export class HttpRequest implements types.HttpRequest {
     async text(): Promise<string> {
         return this.#uReq.text();
     }
+
+    clone(): HttpRequest {
+        const newInit = structuredClone(this.#init);
```
castrodd, line 102: Question: I see that `structuredClone` was supported starting with Node 17.0.0. What if a customer is on an earlier Node version?

test/http/HttpRequest.test.ts
```diff
@@ -11,6 +11,34 @@ import { HttpRequest } from '../../src/http/HttpRequest';
...
+            },
+        });
+        const req2 = req.clone();
+        expect(await req.text()).to.equal('body1');
+        expect(await req2.text()).to.equal('body1');
+        expect(req.headers).to.not.equal(req2.headers);
+        expect(req.headers).to.deep.equal(req2.headers);
+        expect(req.params).to.not.equal(req2.params);
+        expect(req.params).to.deep.equal(req2.params);
+        expect(req.query).to.not.equal(req2.query);
```
castrodd, line 41: Nit: can we have newlines between the different sections (e.g. testing headers, testing params)? It would make it a little easier to follow what's being tested in each pair.

castrodd (approved):

src/http/HttpRequest.ts
ejizba: This version of the package only supports Node.js v18+. The tests run against all supported versions (see [here](https://github.com/Azure/azure-functions-nodejs-library/blob/v3.x/azure-pipelines/test.yml#L19) for v3 and [here](https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/azure-pipelines/test.yml#L17) for v4), so we're protected from these types of concerns

restfulhead:
@ejizba Thank you for adding this. Any chance this could be released to NPM?

ejizba:
Hi @restfulhead our plan is to do a release at the end of January, so should be out pretty soon

Sarabadu:
Hi @ejizba, thank for adding this will help us a lot.
it is possible to have an alpha build to start testing it? we are building some middleware and we didn't find a way to workaround this yet

ejizba:
@Sarabadu You could download the tgz file from our latest v4.x branch build: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=159678&view=artifacts&pathAsName=false&type=publishedArtifacts

(download "drop.zip", unzip that, and from your repo do `npm install <path to azure-functions-4.1.0-alpha.20240111.3.tgz>`)

restfulhead:
> Hi @restfulhead our plan is to do a release at the end of January, so should be out pretty soon

@ejizba Any updates regarding the release?

Btw, the link https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=159678&view=artifacts&pathAsName=false&type=publishedArtifacts takes me to an almost empty page. I don't see any file to download.

ejizba:
Hi we just released v4.2.0 with this change

Issue #210, opened 2024-01-22
Cannot read properties of undefined (reading 'type')


GravlLift:
#### Investigative information

Please provide the following:

- Timestamp: 2024-01-22T22:14:28.4980344Z
- Function App name: AZUD-RMS-FA-01
- Function name(s) (as appropriate): customerLookup
- Invocation ID: 3325acd5-1aad-4923-bffc-0a6fbfe52204
- Region: East US 2

#### Repro steps

Provide the steps required to reproduce the problem:

1. Open `https://azud-rms-fa-01.azurewebsites.net/api/v4/customer?emailAddress=lightspeed.e2e%40gmail.com` in a browser

#### Expected behavior

Provide a description of the expected behavior.

Http request should return valid JSON and 200 response

#### Actual behavior

500 error

#### Related information 

Provide any related information 

* Programming language used: javascript
* Bindings used:

```ts
app.get("customerLookup", {
  authLevel: "anonymous",
  route: `api/v{version:int:min(4):max(${currentMajorVersion})}/customer`,
  handler,
});
```

* Stack Trace:
```
Stack: TypeError: Cannot read properties of undefined (reading 'type')
    at InvocationModel.<anonymous> (/node_modules/@azure/functions/dist/webpack:/@azure/functions/src/InvocationModel.ts:65:63)
    at Generator.next (<anonymous>)
    at <anonymous> (/node_modules/@azure/functions/dist/webpack:/@azure/functions/src/InvocationContext.ts:96:10)
    at new Promise (<anonymous>)
    at __webpack_modules__../src/InvocationModel.ts.__awaiter (/node_modules/@azure/functions/dist/webpack:/@azure/functions/src/InvocationContext.ts:96:10)
    at InvocationModel.getArguments (/node_modules/@azure/functions/dist/webpack:/@azure/functions/src/InvocationModel.ts:46:23)
    at y.<anonymous> (/azure-functions-host/workers/node/dist/src/worker-bundle.js:2:37611)
    at Generator.next (<anonymous>)
    at o (/azure-functions-host/workers/node/dist/src/worker-bundle.js:2:35629)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) 
```

This error is not reproducible locally, only on our Azure instance. While it is reproducible for a single function run, restarting the function will fix the issue.

Relevant code appears to be here:

[...]

PR #211, opened 2024-01-24
Turn on noUncheckedIndexedAccess

ejizba:
https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

Related to https://github.com/Azure/azure-functions-nodejs-library/issues/210, should result in slightly better error message

### DIFF
--- a/scripts/updateVersion.ts
+++ b/scripts/updateVersion.ts
@@ -59,7 +59,7 @@ function validateVersion(): string {
 function getVersion(filePath: string, regex: RegExp): string {
     const fileContents = readFileSync(filePath).toString();
     const match = fileContents.match(regex);
-    if (!match) {
+    if (!match || !match[1]) {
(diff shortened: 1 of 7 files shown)


scripts/updateVersion.ts
```diff
@@ -79,7 +79,7 @@ function updatePackageJsonVersion(cwd: string, newVersion: string) {
 function updateVersionByRegex(filePath: string, regex: RegExp, newVersion: string) {
     const oldFileContents = readFileSync(filePath).toString();
     const match = oldFileContents.match(regex);
-    if (!match) {
+    if (!match || !match[0] || !match[1]) {
         throw new Error(`Failed to find match for "${regex.source}".`);
     }
     const oldLine = match[0];
```
castrodd: We could use array destructuring on line 85:

`const [oldLine, oldVersion] = match;`

castrodd (approved):

Issue #97, opened 2023-05-28
Support streams (http)


haukepribnow:
#### Repro steps

1. Using programming model v4, create an HTTP-triggered function where:
   - The handler function provides a `ReadableStream` in the return object's `body` field.
   - The `ReadableStream` enqueues data with a time delay.
   - The `ReadableStream` continues to enqueue data after the handler function has returned.
   - While enqueuing data, `context.log()` gets called. 

2. Call the function via HTTP.

(I recommend to check the source code below; I guess it's more telling than my description here.)

#### Expected behavior

When the function is called:
- When new data is enqueued, it should directly be sent to the HTTP client as a chunk (using `Transfer-Encoding: chunked`).
- The `context.log()` calls should log "as usual" and should not raise warnings/errors.

#### Actual behavior

- While `Transfer-Encoding: chunked` **is** set in the HTTP response, newly enqueued data is not directly sent when it becomes available. Instead, the completion of the stream is awaited before any data is sent. All data is then sent in a single chunk.
- While the `context.log()` calls (that happen when data is enqueued after the handler function has returned) **do** output the log message, they also output a warning: `Unexpected call to 'log' on the context object after function execution has completed. Please check for asynchronous calls that are not awaited. Function name: streamingTest. Invocation Id: [removed].`

#### Known workarounds

n/a

#### Related information 

* Programming language used: TypeScript

<details>
<summary>Source</summary>

```typescript
import { ReadableStream } from "stream/web";
import { HttpRequest, HttpResponse, InvocationContext, app } from "@azure/functions";

// Test function that returns a ReadableStream as body.
export async function streamingTest(request: HttpRequest, context: InvocationContext): Promise<HttpResponse> {
    context.log(`Http function processed request for url "${request.url}"`);

    const encoder = new TextEncoder();
    let times = 0;

    const stream = new ReadableStream({
        type: 'bytes',
        async start(controller) {
            function onTimer() {
                if (times >= 1000) {
                    controller.close();
                    return;
                }

                context.log(`${times}`);
                // This raises a warning:
                // Unexpected call to 'log' on the context object after function execution
                // has completed. Please check for asynchronous calls that are not awaited.
                // Function name: streamingTest. Invocation Id: [removed].

                times++;
                const data = encoder.encode(`${times}\r\n`);
                controller.enqueue(data);
                setTimeout(onTimer, 10);
            }
            onTimer();
        } 
    }, {
       highWaterMark: 1 
    });

[...]

PR #212, opened 2024-01-30
Http stream support

ejizba:
HTTP stream support should require minimal changes by the user. The only thing _required_ is turning on this flag:
```typescript
import { app } from '@azure/functions';

app.setup({ enableHttpStream: true });
```
Otherwise it's up to them to use the existing HTTP request/response types as streams instead of buffers.

I've added end-to-end tests here which can be used as sample code for now: https://github.com/Azure/azure-functions-nodejs-e2e-tests/pull/32

Lastly, the only known problem is that the `request.params` object will not be populated. Plenty of scenarios don't need this and users should be able to workaround the problem by parsing the url themselves, so I'd rather get this moving while we figure that specific feature out offline.

Depends on [v3.9.0 of the worker](https://github.com/Azure/azure-functions-nodejs-worker/releases/tag/v3.9.0) which has finished rolling out in Azure with host v4.28, but we don't have an official core tools release yet.

Related to https://github.com/Azure/azure-functions-nodejs-library/issues/97
Related to https://github.com/Azure/azure-functions-nodejs-library/issues/171

### DIFF
--- a/src/InvocationModel.ts
+++ b/src/InvocationModel.ts
@@ -136,12 +136,12 @@ export class InvocationModel implements coreTypes.InvocationModel {
     }
 
     async #convertOutput(
-        invocId: string,
+        invocationId: string,
         binding: RpcBindingInfo,
         value: unknown
     ): Promise<RpcTypedData | null | undefined> {
         if (binding.type?.toLowerCase() === 'http') {
-            return toRpcHttp(invocId, value);
+            return toRpcHttp(invocationId, value);
         } else {
             return toRpcTypedData(value);
         }
--- a/src/InvocationModel.ts
+++ b/src/InvocationModel.ts
@@ -20,7 +20,10 @@ import { fromRpcTypedData } from './converters/fromRpcTypedData';
 import { toCamelCaseValue } from './converters/toCamelCase';
 import { toRpcHttp } from './converters/toRpcHttp';
 import { toRpcTypedData } from './converters/toRpcTypedData';
+import { waitForRequest } from './http/httpProxy';
+import { HttpRequest } from './http/HttpRequest';
 import { InvocationContext } from './InvocationContext';
+import { isHttpStreamEnabled } from './setup';
 import { isHttpTrigger, isTimerTrigger, isTrigger } from './utils/isTrigger';
 import { isDefined, nonNullProp, nonNullValue } from './utils/nonNull';
 
@@ -60,9 +63,17 @@ export class InvocationModel implements coreTypes.InvocationModel {
         if (req.inputData) {
             for (const binding of req.inputData) {
                 const bindingName = nonNullProp(binding, 'name');
-                let input: unknown = fromRpcTypedData(binding.data);
 
                 const bindingType = nonNullProp(this.#bindings, bindingName).type;
+
+                let input: unknown;
+                if (isHttpTrigger(bindingType) && isHttpStreamEnabled()) {
+                    const proxyRequest = await waitForRequest(this.#coreCtx.invocationId);
+                    input = new HttpRequest({ ...binding.data?.http, proxyRequest });
+                } else {
+                    input = fromRpcTypedData(binding.data);
+                }
+
                 if (isTimerTrigger(bindingType)) {
                     input = toCamelCaseValue(input);
                 }
@@ -98,10 +109,14 @@ export class InvocationModel implements coreTypes.InvocationModel {
         for (const [name, binding] of Object.entries(this.#bindings)) {
             if (binding.direction === 'out') {
                 if (name === returnBindingKey) {
-                    response.returnValue = await this.#convertOutput(binding, result);
+                    response.returnValue = await this.#convertOutput(context.invocationId, binding, result);
                     usedReturnValue = true;
                 } else {
-                    const outputValue = await this.#convertOutput(binding, context.extraOutputs.get(name));
+                    const outputValue = await this.#convertOutput(
+                        context.invocationId,
+                        binding,
+                        context.extraOutputs.get(name)
+                    );
                     if (isDefined(outputValue)) {
                         response.outputData.push({ name, data: outputValue });
                     }
@@ -120,9 +135,13 @@ export class InvocationModel implements coreTypes.InvocationModel {
         return response;
     }
 
-    async #convertOutput(binding: RpcBindingInfo, value: unknown): Promise<RpcTypedData | null | undefined> {
+    async #convertOutput(
+        invocId: string,
+        binding: RpcBindingInfo,
+        value: unknown
+    ): Promise<RpcTypedData | null | undefined> {
         if (binding.type?.toLowerCase() === 'http') {
-            return toRpcHttp(value);
+            return toRpcHttp(invocId, value);
         } else {
             return toRpcTypedData(value);
         }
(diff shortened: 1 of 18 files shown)


src/InvocationModel.ts
```diff
@@ -120,9 +135,13 @@ export class InvocationModel implements coreTypes.InvocationModel {
         return response;
     }
 
-    async #convertOutput(binding: RpcBindingInfo, value: unknown): Promise<RpcTypedData | null | undefined> {
+    async #convertOutput(
+        invocId: string,
```
castrodd: Nit: We use `invocationId` pretty regularly elsewhere in the code. I think it's better to use the full name to remain consistent.

src/converters/toRpcHttp.ts
```diff
@@ -3,11 +3,13 @@
...
 import { RpcHttpData, RpcTypedData } from '@azure/functions-core';
 import { AzFuncSystemError } from '../errors';
+import { sendResponse } from '../http/httpProxy';
 import { HttpResponse } from '../http/HttpResponse';
+import { isHttpStreamEnabled } from '../setup';
 import { toRpcHttpCookie } from './toRpcHttpCookie';
 import { toRpcTypedData } from './toRpcTypedData';
 
-export async function toRpcHttp(data: unknown): Promise<RpcTypedData | null | undefined> {
+export async function toRpcHttp(invocId: string, data: unknown): Promise<RpcTypedData | null | undefined> {
```
castrodd: Nit: same suggestion as below. I think using the full name for `invocationId` is clearer and more consistent.

src/converters/toRpcHttp.ts
```diff
@@ -17,6 +19,11 @@ export async function toRpcHttp(data: unknown): Promise<RpcTypedData | null | un
     }
 
     const response = data instanceof HttpResponse ? data : new HttpResponse(data);
+    if (isHttpStreamEnabled()) {
+        await sendResponse(invocId, response);
+        return undefined;
```
castrodd: Question: why do we need to return `undefined` when streaming is enabled?

src/http/HttpRequest.ts
```diff
@@ -27,9 +31,16 @@ export class HttpRequest implements types.HttpRequest {
...
-        if (init.undiciRequest) {
-            this.#uReq = init.undiciRequest;
+        if (init.proxyRequest) {
+            [this.#uReq, this.query, this.params] = this.#initStreamRequest(init);
         } else {
+            [this.#uReq, this.query, this.params] = this.#initInMemoryRequest(init);
+        }
+    }
+
+    #initInMemoryRequest(init: InternalHttpRequestInit): [uRequest, URLSearchParams, HttpRequestParams] {
```
castrodd: What do you think about changing the return type from an array to a proper type? We can still use destructuring in the constructor, but this way we have a cleaner function signature and we signal to consumers more directly why these types are being bundled together.

src/http/HttpRequest.ts
```diff
@@ -39,15 +50,51 @@ export class HttpRequest implements types.HttpRequest {
...
+        const query = new URLSearchParams(fromNullableMapping(init.nullableQuery, init.query));
+        const params = fromNullableMapping(init.nullableParams, init.params);
+
+        return [uReq, query, params];
+    }
+
+    #initStreamRequest(init: InternalHttpRequestInit): [uRequest, URLSearchParams, HttpRequestParams] {
+        const proxyReq = nonNullProp(init, 'proxyRequest');
+
+        const hostHeaderName = 'x-forwarded-host';
```
castrodd, line 71: I would encapsulate lines 69-75 in a private function (`constructURL()`?) and then line 76 could just be `const url = constructURL(proxyReq);`.

src/http/HttpRequest.ts
```diff
@@ -39,15 +50,51 @@ export class HttpRequest implements types.HttpRequest {
...
+        const protoHeaderName = 'x-forwarded-proto';
+        const host = proxyReq.headers[hostHeaderName];
+        const proto = proxyReq.headers[protoHeaderName];
+        if (typeof host !== 'string' || typeof proto !== 'string') {
+            throw new AzFuncSystemError(`Expected headers "${hostHeaderName}" and "${protoHeaderName}" to be set.`);
+        }
+        const url = `${proto}://${host}${nonNullProp(proxyReq, 'url')}`;
+
+        let uReq = init.undiciRequest;
+        if (!uReq) {
```
castrodd, line 81: Similar idea to below. We can abstract away the code to set uReq and just have `const uReq =  constructUndiciReq(init.undiciReques);`.

This way `#initStreamRequest` itself consists of merely 5 or 6 lines and we can clearly see its key components in a glance.

src/http/httpProxy.ts
```diff
@@ -0,0 +1,119 @@
...
+    proxyRes.end();
+}
+
+function setCookies(userRes: HttpResponse, proxyRes: http.ServerResponse): void {
+    const serializedCookies: string[] = userRes.cookies.map((c) => {
+        let sameSite: true | false | 'lax' | 'strict' | 'none' | undefined;
+        switch (c.sameSite) {
+            case 'Lax':
+                sameSite = 'lax';
+                break;
```
castrodd, line 63: We could also do something like:

```
sameSite = typeof c.sameSite === "string" ? c.sameSite.toLowerCase() : c.sameSite;
```

If we don't want to check against `string`, we could create a union type of lax, strict, and none.

src/InvocationModel.ts
ejizba: Fixed

src/converters/toRpcHttp.ts
ejizba: Fixed

src/converters/toRpcHttp.ts
ejizba: If you're asking why do we do `return undefined;` vs. `return;` I don't have a reason and fixed to do the shorter option.

If you're asking why do we return at a more general level, it's because the http response is handled over the http proxy when streaming instead of over rpc. Tweaked the naming and added a comment in case that helps

src/http/HttpRequest.ts
ejizba: Not entirely sure, but I think I did what you suggested

src/http/HttpRequest.ts
ejizba: I would normally encapsulate something when either (a) it needs to be used in two places or (b) the method is getting too long. This code is only used once and I guess I feel like 20-30 lines is perfectly fine

Otherwise I find it more readable to just have one method and use spacing/comments to delineate sections if necessary. It's personal preference, but I just don't like navigating a whole call stack of methods if I want to see the full code for a (reasonably sized) method. Those methods could get out of order in the file, have other methods with confusingly similar names, and make it more difficult to refactor in the future.

src/http/HttpRequest.ts
ejizba: see other comment

src/http/httpProxy.ts
ejizba: If I do that I get a TypeScript error:

> Type 'string | undefined' is not assignable to type 'boolean | "lax" | "strict" | "none" | undefined'.
  Type 'string' is not assignable to type 'boolean | "lax" | "strict" | "none" | undefined'.

The only reason I have to do this is because rpc and the cookie parser both use string constant union types, but unfortunately they handle the casing differently. If I use `toLowerCase`, I lose type checking because it's converted to a generic string instead of the string constant union type

castrodd (approved):

amcdnl:
Was this released in the 4.2.0 release 2 weeks ago? If not, is there an ETA?

ejizba:
No it wasn't, please see https://github.com/Azure/azure-functions-nodejs-library/issues/97#issuecomment-1915798480

PR #213, opened 2024-01-30
Make error message even better for missing binding

ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-library/issues/210

### DIFF
--- a/src/InvocationModel.ts
+++ b/src/InvocationModel.ts
@@ -20,6 +20,7 @@ import { fromRpcTypedData } from './converters/fromRpcTypedData';
 import { toCamelCaseValue } from './converters/toCamelCase';
 import { toRpcHttp } from './converters/toRpcHttp';
 import { toRpcTypedData } from './converters/toRpcTypedData';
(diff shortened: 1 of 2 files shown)


src/InvocationModel.ts
```diff
@@ -62,7 +63,13 @@ export class InvocationModel implements coreTypes.InvocationModel {
                 const bindingName = nonNullProp(binding, 'name');
                 let input: unknown = fromRpcTypedData(binding.data);
 
-                const bindingType = nonNullProp(this.#bindings, bindingName).type;
+                const rpcBinding = this.#bindings[bindingName];
+                if (!rpcBinding) {
+                    throw new AzFuncSystemError(
+                        `Failed to find binding "${bindingName}" in bindings "${Object.keys(this.#bindings).join()}".`
```
castrodd: Should we consider `.join(", ")` instead of `.join()` so that there is some spacing in the list of keys? Right now, we log the bindings without any separation: `"httpTrigger1,$return"`

castrodd (approved):

PR #216, opened 2024-02-02
Fix out-of-sync binding names causing null ref error

### LINKED ISSUE #210: Cannot read properties of undefined (reading 'type')

ejizba:
Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/210

The logs from the user indicate a mismatch in the binding name between the host and worker ("httpTrigger17" vs "httpTrigger19"). I suspect our logic in "addBindingName" isn't reliable enough at returning the same binding name. When I first adjusted the logic in "addBindingName" (see [the original issue](https://github.com/Azure/azure-functions-nodejs-worker/issues/638)) I thought the problem only repro-ed with multiple workers in parallel, but now I think it can also happen with multiple workers in serial (aka a worker restarts, but the host stays the same). Fwiw, I checked telemetry and only a handful of apps are hitting this issue, but I definitely don't want anyone to hit this at all.

Anyways, my new solution is to create a hash of the binding configuration, which should reliably create the same binding name each time and also be unique within the function (we need name uniqueness at the function level, but not the app level). The only way we could get conflicting names is if they register bindings with the exact same config within a single function, but I have no idea why someone would do that. Regardless, I added an error up front so that it will fail-fast if they're in that situation and be easy for them to fix.

Lastly, I split this PR into two commits for easier review:
- The first moves code to a new file with no functional changes: https://github.com/Azure/azure-functions-nodejs-library/commit/2e24f147ac5a04243c2f5502f48a09aa138da5bf
- The second actually fixes the issue: https://github.com/Azure/azure-functions-nodejs-library/commit/2f08eb60776485200d678a5725667daca5baeb14

### DIFF
--- a/src/converters/toCoreFunctionMetadata.ts
+++ b/src/converters/toCoreFunctionMetadata.ts
@@ -1,0 +1,63 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { ExponentialBackoffRetryOptions, FixedDelayRetryOptions, GenericFunctionOptions } from '@azure/functions';
+import * as coreTypes from '@azure/functions-core';
+import { returnBindingKey } from '../constants';
+import { isTrigger } from '../utils/isTrigger';
+import { toRpcDuration } from './toRpcDuration';
+
+export function toCoreFunctionMetadata(name: string, options: GenericFunctionOptions): coreTypes.FunctionMetadata {
+    const bindings: Record<string, coreTypes.RpcBindingInfo> = {};
+
+    const trigger = options.trigger;
+    bindings[trigger.name] = {
+        ...trigger,
+        direction: 'in',
+        type: isTrigger(trigger.type) ? trigger.type : trigger.type + 'Trigger',
+    };
+
+    if (options.extraInputs) {
+        for (const input of options.extraInputs) {
+            bindings[input.name] = {
+                ...input,
+                direction: 'in',
+            };
+        }
+    }
+
+    if (options.return) {
+        bindings[returnBindingKey] = {
+            ...options.return,
+            direction: 'out',
+        };
+    }
+
+    if (options.extraOutputs) {
+        for (const output of options.extraOutputs) {
+            bindings[output.name] = {
+                ...output,
+                direction: 'out',
+            };
+        }
+    }
+
+    let retryOptions: coreTypes.RpcRetryOptions | undefined;
+    if (options.retry) {
+        retryOptions = {
+            ...options.retry,
+            retryStrategy: options.retry.strategy,
+            delayInterval: toRpcDuration((<FixedDelayRetryOptions>options.retry).delayInterval, 'retry.delayInterval'),
+            maximumInterval: toRpcDuration(
+                (<ExponentialBackoffRetryOptions>options.retry).maximumInterval,
+                'retry.maximumInterval'
+            ),
+            minimumInterval: toRpcDuration(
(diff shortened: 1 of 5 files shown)


src/converters/toCoreFunctionMetadata.ts
```diff
@@ -0,0 +1,76 @@
...
+                direction: 'out',
+            };
+            bindingNames.push(output.name);
+        }
+    }
+
+    const dupeBindings = bindingNames.filter((v, i) => bindingNames.indexOf(v) !== i);
+    if (dupeBindings.length > 0) {
+        throw new AzFuncSystemError(
+            `Duplicate bindings found for function "${name}". Remove a duplicate binding or manually specify the "name" property to make it unique.`
```
castrodd, line 54: Should we capture the duplicate binding's name and log in this error message? Or is the number of bindings so small that the user can figure it out without the information?

castrodd (approved):

src/converters/toCoreFunctionMetadata.ts
ejizba: The name is auto-generated, so it probably own't mean much to the user. And yeah I would expect most functions have 1-3 bindings, with anything more than 5 to be abnormally high

PR #222, opened 2024-02-16
Add aka.ms link to stream flag

ejizba:


castrodd (approved):

types/setup.d.ts
```diff
@@ -3,7 +3,8 @@
 
 export interface SetupOptions {
     /**
-     * PREVIEW: Stream http requests and responses instead of loading entire body in memory
+     * PREVIEW: Stream http requests and responses instead of loading entire body in memory.
+     * [Learn more here](https://aka.ms/AzFuncNodeHttpStreams)
```
castrodd, line 7: Since this isn't a markdown file, should we skip the link formatting?
```
Learn more here: https://aka.ms/AzFuncNodeHttpStreams
```

ejizba: Screenshot 2024-02-16 at 3 41 11 PM

VS Code actually supports markdown links for their intellisense, which is why I do it. Probably not every tool supports it and obviously someone could look at the types file directly, but I feel like even with the braces/parentheses it's still pretty readable, but let me know if you disagree

Issue #168, opened 2023-04-10
Support Arrays In Query String


PCOffline:
There is a convention for the syntax to pass an array in query parameters:

1. Passing the same parameter multiple times:
`?foo=bar&foo=qux`
2. Passing the same parameter multiple times with a `[]` suffix:
`?foo[]=bar&foo[]=qux` - URL encoded as `?foo%5B%5D=bar&foo%5B%5D=qux`
3. Passing the same parameter multiple times with an explicit index in the `[]` suffix:
`?foo[0]=bar&foo[1]=qux` - URL encoded as `?foo%5B0%5D=bar&foo%5B1%5D=qux`
or, to demonstrate the functionality of the explicit index:
`?foo[1]=qux&foo[0]=bar` - URL encoded as `?foo%5B1%5D=qux&foo%5B0%5D=bar`
4. Passing a comma-separated list:
`?foo=bar,qux`

Sadly, there is no unified standard or W3C spec, and every library and framework handles these cases differently.
As of today, none of these options will produce an array in Azure Functions.

The change I propose is a **breaking change**, so perhaps an opt-in should be added, or it could be merged to version 5 if such a version is upcoming.

#### Investigative information
N/A

#### Repro steps
1. Create a sample HTTP function that prints its query parameters
2. Send a request that uses one of the syntaxes for arrays in query params (for example `?foo[]=bar&foo[]=qux`)

#### Expected behaviour
We should receive an object that contains the parameter as the field and an array as the value.
For example, when sending `?foo[]=bar&foo[]=qux` we should get `{ foo: ["bar", "qux"] } as our query.

#### Actual behaviour
```js
"?foo=bar&foo=qux"         => { "foo": "bar,qux" }
"?foo[]=bar&foo[]=qux"     => { "foo[]": "bar,qux" }
"?foo[1]=qux&foo[0]=bar"   => { "foo[0]": "bar", "foo[1]": "qux" }
"?foo=bar,qux"             => { "foo": "bar,qux" }
```

#### Known workarounds
A workaround is available if we utilize the `qs` (short for query string) or `query-string` libraries and parse the query params with the next code:
**Be advised this code is overly simplified and does not cover edge cases**.
```js
import type { AzureFunction, Context, HttpRequest } from "@azure/functions";
import qs from "qs";

const httpTrigger: AzureFunction = async (context: Context, req: HttpRequest): Promise<void> => {
  const query = qs.parse(req.url.split("?")[1]);
[...]

PR #238, opened 2024-03-15
Support params when streaming

### LINKED ISSUE #229: Params are missing when using http streams
When using [http streams](https://aka.ms/AzFuncNodeHttpStreams), the `request.params` object will be an empty object. The suggested workaround is to use a package like [path-to-regexp](https://www.npmjs.com/package/path-to-regexp) to parse the `request.url` property. Keep in mind that Azure Functions only supports [ASP.NET Core route templates](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing#route-templates) (i.e. `user/{name}`) when registering your function, so you will likely need to use a different syntax (i.e. `/user/:name`) when parsing the url. Here is example code using the suggested workaround:

```typescript
import { app, HttpRequest, HttpResponseInit, InvocationContext } from '@azure/functions';
import { match } from 'path-to-regexp';

export async function httpTrigger1(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
    const parseRoute = match('/api/user/:name');
    const route = parseRoute(new URL(request.url).pathname);
    const params = route ? (route.params as Record<string, string>) : {};

    const name = params.name || 'world';

    return { body: `Hello, ${name}!` };
}

app.http('httpTrigger1', {
    methods: ['GET', 'POST'],
    route: 'user/{name}',
    authLevel: 'anonymous',
    handler: httpTrigger1,
});
```

Related to https://github.com/Azure/azure-functions-host/issues/9840

ejizba:
And fix headers to match original user's request. Both of these are made possible because of the trigger metadata property. I tried using query from here as well, but it ended up being the worst of both worlds - it doesn't fix the query array bug (https://github.com/Azure/azure-functions-nodejs-library/issues/168) but it also doesn't match the old non-stream behavior. So I left our current logic which fixes the query array bug.

I did a small refactor in the first commit and second commit is the actual change.

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/229

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,7 @@
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-namespace": "off",
-        "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
+        "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_", "ignoreRestSiblings": true }],
         "prefer-const": ["error", { "destructuring": "all" }],
         "@typescript-eslint/explicit-member-accessibility": [
             "error",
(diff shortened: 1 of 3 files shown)


castrodd (approved):

PCOffline:
Hey, does this mean that the latest 4.x version support arrays in query params out of the box?

ejizba:
@PCOffline only if you have the preview stream feature enabled. More info here: https://aka.ms/AzFuncNodeHttpStreams

PR #240, opened 2024-03-25
Add support for sql trigger

### LINKED ISSUE #88: Add SQL trigger support to v4 model
More information about SQL Bindings can be found [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql?tabs=in-process%2Cextensionv4&pivots=programming-language-csharp).

ejizba:
We already have sql input/output bindings and I was just waiting for the sql trigger to GA before I added it (which happened somewhat recently). SQL trigger docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-azure-sql-trigger?tabs=isolated-process%2Cportal&pivots=programming-language-javascript)

See example code in the e2e test PR here: https://github.com/Azure/azure-functions-nodejs-e2e-tests/pull/36

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/88

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,3 +15,9 @@ export { InvocationContext } from './InvocationContext';
 export * as output from './output';
 export * as trigger from './trigger';
 export { Disposable } from './utils/Disposable';
+
+export enum SqlChangeOperation {
+    Insert = 0,
+    Update = 1,
+    Delete = 2,
+}
--- a/types/sql.d.ts
+++ b/types/sql.d.ts
@@ -1,7 +1,40 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-import { FunctionInput, FunctionOutput } from './index';
+import { FunctionInput, FunctionOptions, FunctionOutput, FunctionResult, FunctionTrigger } from './index';
+import { InvocationContext } from './InvocationContext';
+
+export type SqlHandler = (changes: SqlChange[], context: InvocationContext) => FunctionResult;
+
+export interface SqlFunctionOptions extends SqlTriggerOptions, Partial<FunctionOptions> {
+    handler: SqlHandler;
+
+    trigger?: SqlTrigger;
+}
+
+export interface SqlTriggerOptions {
+    /**
+     * The name of the table monitored by the trigger.
+     */
+    tableName: string;
+
+    /**
+     * An app setting (or environment variable) with the connection string for the database containing the table monitored for changes
+     */
+    connectionStringSetting: string;
+}
+export type SqlTrigger = FunctionTrigger & SqlTriggerOptions;
+
+export interface SqlChange {
+    Item: unknown;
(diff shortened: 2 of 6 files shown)


types/sql.d.ts
```diff
@@ -1,7 +1,40 @@
...
+
+    /**
+     * An app setting (or environment variable) with the connection string for the database containing the table monitored for changes
+     */
+    connectionStringSetting: string;
+}
+export type SqlTrigger = FunctionTrigger & SqlTriggerOptions;
+
+export interface SqlChange {
+    Item: unknown;
```
ejizba, line 29: I wish these were lowercase, and I'm also not a fan of "Operation" using an enum, but this is what the SQL trigger binding gives us, so that's what the types have to reflect (unless we decide to automatically convert the trigger input, but we don't do that for any other trigger input and I didn't think it was worth it)

src/index.ts
```diff
@@ -15,3 +15,9 @@ export { InvocationContext } from './InvocationContext';
 export * as output from './output';
 export * as trigger from './trigger';
 export { Disposable } from './utils/Disposable';
+
+export enum SqlChangeOperation {
```
castrodd, line 19: Have we used enums before in this library? I couldn't find any other examples.

We could use a union type instead: `type SqlChangeOperation = 'insert' | 'update' | 'delete'`

Or, if we want to keep the enum, I think it's preferable to use strings for type safety and clarity:

```
export enum SqlChangeOperation {
    Insert = 'insert',
    Update = 'update',
    Delete = 'delete',
}

```

What do you think?

types/sql.d.ts
```diff
@@ -1,7 +1,40 @@
...
+}
+
+export interface SqlTriggerOptions {
+    /**
+     * The name of the table monitored by the trigger.
+     */
+    tableName: string;
+
+    /**
+     * An app setting (or environment variable) with the connection string for the database containing the table monitored for changes
```
castrodd, line 22: For consistency, should we just use the description from our docs:

```
The name of an app setting that contains the connection string for the database against which the query or stored procedure is being executed.
```

castrodd (approved):

src/index.ts
ejizba: See my other comment. I don't like enums (which is why you can't find other examples), but the sql extension is passing us 0, 1, and 2 for these values so I don't see an easy way around this

types/sql.d.ts
ejizba: The existing comment is consistent with our other binding descriptions in this repo (cosmos, service bus, etc.), and I'd rather maintain consistency within the repo

src/index.ts
ejizba: I could do a union type, but it would have to be `type SqlChangeOperation = 0 | 1 | 2;` because again the sql extension is giving us these as numbers not strings. And I figured if we have to use numbers, then the enum is better

Issue #245, opened 2024-04-11
Support open telemetry


ejizba:
~We plan to create a new package `@opentelemetry/instrumentation-azure-functions` that allows Open Telemetry logs, traces, and metrics to be sent from your Node.js code and properly correlated with all Azure Functions Host data for your invocations. The package will be included in [@opentelemetry/auto-instrumentations-node](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node), so that you can get all instrumentations in one place.~ Edit: due to https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2259, we shifted to an `@azure` package instead.

Until that package is released, you can replicate this behavior with the following code: https://github.com/ejizba/azure-functions-nodejs-opentelemetry/blob/main/src/index.ts#L24

Related to https://github.com/Azure/azure-functions-host/issues/9273

ejizba:
First version of the functions open telemetry package has been released: https://www.npmjs.com/package/@azure/functions-opentelemetry-instrumentation

GitHub repo here: https://github.com/Azure/azure-functions-nodejs-opentelemetry

I also updated our docs here: https://learn.microsoft.com/en-us/azure/azure-functions/opentelemetry-howto?tabs=otlp-export&pivots=programming-language-typescript

PR #255, opened 2024-05-08
Add ability to specify capabilities in setup

ejizba:
and clarify that setup options are merged instead of replaced. Otherwise there would be no way for multiple places (i.e. the user's app and an otel package) to adjust setup config.

Related to otel work https://github.com/Azure/azure-functions-nodejs-library/issues/245, we will need to do something like this:
```typescript
app.setup({
    capabilities: {
        WorkerOpenTelemetryEnabled: true,
    },
});
```

### DIFF
--- a/src/ProgrammingModel.ts
+++ b/src/ProgrammingModel.ts
@@ -16,16 +16,16 @@ export class ProgrammingModel implements coreTypes.ProgrammingModel {
         return new InvocationModel(coreCtx);
     }
 
-    async getCapabilities(capabilities: WorkerCapabilities): Promise<WorkerCapabilities> {
+    async getCapabilities(workerCapabilities: WorkerCapabilities): Promise<WorkerCapabilities> {
         lockSetup();
 
         if (enableHttpStream) {
             const httpUri = await setupHttpProxy();
-            capabilities.HttpUri = httpUri;
+            workerCapabilities.HttpUri = httpUri;
         }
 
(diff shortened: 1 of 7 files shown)


src/ProgrammingModel.ts
```diff
@@ -19,11 +19,13 @@ export class ProgrammingModel implements coreTypes.ProgrammingModel {
     async getCapabilities(capabilities: WorkerCapabilities): Promise<WorkerCapabilities> {
         lockSetup();
 
-        if (isHttpStreamEnabled()) {
+        if (enableHttpStream) {
             const httpUri = await setupHttpProxy();
             capabilities.HttpUri = httpUri;
         }
 
+        Object.assign(capabilities, libraryCapabilities);
```
castrodd: Should we also change the name of this parameter from just `capabilities` to `workerCapabilities`? Just so we are clearly distinguishing the two different capabilities.

test/setup.test.ts
```diff
@@ -0,0 +1,46 @@
...
+        setup({ enableHttpStream: true });
+        expect(enableHttpStream).to.equal(true);
+
+        // don't change if not explicitly set
+        setup({});
+        expect(enableHttpStream).to.equal(true);
+        setup({ capabilities: {} });
+        expect(enableHttpStream).to.equal(true);
+
+        // set to true
```
castrodd: Should the comment on line 23 be 'set to false'?

castrodd (approved):

PR #275, opened 2024-06-21
Run builds nightly

ejizba:


azure-pipelines/official-build.yml
```diff
@@ -12,6 +12,14 @@ trigger:
 # CI only, does not trigger on PRs.
 pr: none
 
+schedules:
+    - cron: '30 10 * * *'
+      displayName: Nightly at 3:30 AM PT
```
castrodd: Same comment as here: https://github.com/Azure/azure-functions-nodejs-worker/pull/751#discussion_r1651290034

But, also, do we want to run all these builds at the same time? Or should we spread them out?

castrodd (approved):

PR #286, opened 2024-08-02
Http: Accept an optional type in json functions

orgads:


ejizba:
Hi @orgads, thanks for the contribution! We generally just follow the community for these http types and none of them do this. Here's just one comment that explains why: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36868#issuecomment-512989650

PR #288, opened 2024-08-09
Fix undefined params when http streaming is on

### LINKED ISSUE #285: v4 model non-string route parameters are undefined
We are migrating from v3 to v4 currently and have reached a critical problem. It seems like the routing does not work with numbers/int. 

I took the basic example for Docs: https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cfunctionsv2&pivots=programming-language-typescript#customize-the-http-endpoint

We can see that the category is resolving perfectly fine, but the id is undefined. 
```typescript
import { app, HttpRequest, HttpResponseInit, InvocationContext } from '@azure/functions';

export async function httpTrigger1(request: HttpRequest, context: InvocationContext): Promise<HttpResponseInit> {
  console.log(request.params); // { category: 'books' }
  const category = request.params.category; // books
  const id = request.params.id; // undefined

  return { body: `Category: ${category}, ID: ${id}` }; // Category: books, ID: undefined
}

app.http('httpTrigger1', {
    methods: ['GET'],
    authLevel: 'anonymous',
    route: 'products/{category:alpha}/{id:int?}',
    handler: httpTrigger1,
});
```

No matter if I remove the route constraint it still is undefined. 

System: 
```
Azure Functions Core Tools
Core Tools Version:       4.0.5858 Commit hash: N/A +f9ca1cc1f385317a558bc35b3a3fc96af22665e4 (64-bit)
Function Runtime Version: 4.834.3.22875
```

ejizba:
When the http stream feature is turned on, some params are not being set. This is because `fromRpcTypedData` tries to JSON parse strings and will return values that are objects/numbers/etc which we don't want. Fix was to bypass `fromRpcTypedData` and return the string directly.

Fixes https://github.com/Azure/azure-functions-nodejs-library/issues/285

### DIFF
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -153,9 +153,8 @@ export function createStreamRequest(
 
     const params: Record<string, string> = {};
     for (const [key, rpcValue] of Object.entries(rpcParams)) {
-        const value = fromRpcTypedData(rpcValue);
-        if (typeof value === 'string') {
-            params[key] = value;
+        if (isDefined(rpcValue.string)) {
+            params[key] = rpcValue.string;
         }
     }
 


ejizba:
ping @hallvictoria @gavin-aguiar

Also for the matching test PR here: https://github.com/Azure/azure-functions-nodejs-e2e-tests/pull/48/files

gavin-aguiar (approved):

PR #303, opened 2024-10-08
Upgrade 'typescript' to v5

### LINKED ISSUE #269: Officially support TypeScript 5
TypeScript v5 was released in March 2023. It's possible customers already got our package working with TS v5, but we need to investigate and officially add support.

Ideally we can support TypeScript v4 and v5 at the same time, but if not we should switch to TS v5 when we do the next major version of the `@azure/functions` package.

We have unit tests to cover TypeScript versions, which would need some updating:
- Add TS v5 in package.json similar to how we have "typescript4": `"typescript5": "npm:typescript@~5.0.0"`
  - NOTE: There's a difference between the TS version used to build our code (the default) and these ones which are only used for testing. The test versions should be on minimum supported major versions for ideal validation, which is why they use "~" instead of "^"
- Add '5' to the array of versions in the test [here](https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/test/Types.test.ts#L10)
- Add several default functions to [this file](https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/test/types/index.test.ts). You can compare to the v3.x branch to see what it should look like. Apparently we removed this temporarily for the new model and never added it back

fseitun:
Closes #269 

Upgrades typescript to V5

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
         "@types/long": "^4.0.2",
         "@types/minimist": "^1.2.2",
         "@types/mocha": "^9.1.1",
-        "@types/node": "^18.0.0",
+        "@types/node": "^22.7.5",
         "@types/semver": "^7.3.9",
         "@typescript-eslint/eslint-plugin": "^5.12.1",
         "@typescript-eslint/parser": "^5.12.1",
@@ -61,7 +61,7 @@
         "chai-as-promised": "^7.1.1",
         "eslint": "^7.32.0",
         "eslint-config-prettier": "^8.3.0",
-        "eslint-plugin-deprecation": "^1.3.2",
+        "eslint-plugin-deprecation": "^2.0.0",
         "eslint-plugin-header": "^3.1.1",
         "eslint-plugin-import": "^2.29.0",
         "eslint-plugin-prettier": "^4.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -53,15 +53,15 @@
         "@types/long": "^4.0.2",
         "@types/minimist": "^1.2.2",
         "@types/mocha": "^9.1.1",
-        "@types/node": "^18.0.0",
+        "@types/node": "^22.7.5",
```
ejizba, line 56: You can bump this to a higher minor or patch version of 18.x, but we need the major version to stay 18 as that is our minimum supported node.js version

package.json
```diff
@@ -78,8 +78,8 @@
         "semver": "^7.3.5",
         "ts-loader": "^9.3.1",
         "ts-node": "^3.3.0",
-        "typescript": "^4.5.5",
-        "typescript4": "npm:typescript@~4.0.0",
+        "typescript": "^5.0.0",
```
ejizba, line 81: Hi @fseitun, we can't bump TypeScript across major versions without doing a new major version ourselves, which we're not ready to do at this time. The only way we would support TypeScript 5 is if we can support both 4 and 5 at the same time, which I honestly just haven't tested yet.

Here's how you would do it:
1. Instead of replacing `typescript4` in this file, add another line for `typescript5`
2. In the test file, add '5' to the array instead of replacing '4'

Then it's a matter of figuring out how to get the test for both 4 and 5 passing at the same time, which I anticipate is the actually interesting part of this.

ejizba (changes requested):

fseitun:
Thanks for the feedback! I'm pausing this momentarily, I will try to come back as soon as I find time.

PR #315, opened 2024-11-14
Add port validation checks for HTTP streaming

hallvictoria:
For HTTP streaming, the port is selected by the platform. However, there was no logic that checks if the selected port is valid and available. In the case where the OS is Windows and VNET integration is enabled, the platform defaults to port 0. Port 0 isn't valid, so any proxy requests fail, and the function execution times out.

This fix adds validation on the selected port value. If the port value is 0, the worker will search for and bind to an available port. It first starts searching at port 55000, and if that port isn't available, it will pick a random port between 55000 and 55025 (max port value) and search again. It continues this process until a port is found or throws an error if none are found.

There are no behavioral changes if the selected port value is not 0. The only scenario this change affects then is for Windows + VNET integration enabled.

### DIFF
--- a/src/http/httpProxy.ts
+++ b/src/http/httpProxy.ts
@@ -17,16 +17,13 @@ const invocRequestEmitter = new EventEmitter();
 
 export async function waitForProxyRequest(invocationId: string): Promise<http.IncomingMessage> {
     return new Promise((resolve, _reject) => {
-        workerSystemLog('debug', 'Waiting for proxy request', { invocationId });
         const req = requests[invocationId];
         if (req) {
             resolve(req);
-            workerSystemLog('debug', 'Proxy request found', { invocationId });
             delete requests[invocationId];
         } else {
             invocRequestEmitter.once(invocationId, () => {
                 const req = requests[invocationId];
-                workerSystemLog('debug', 'Proxy request else statement');
                 if (req) {
                     resolve(req);
                     delete requests[invocationId];
@@ -38,13 +35,11 @@ export async function waitForProxyRequest(invocationId: string): Promise<http.In
 
 const invocationIdHeader = 'x-ms-invocation-id';
 export async function sendProxyResponse(invocationId: string, userRes: HttpResponse): Promise<void> {
-    workerSystemLog('debug', 'Sending proxy response', { invocationId });
     const proxyRes = nonNullProp(responses, invocationId);
     delete responses[invocationId];
     for (const [key, val] of userRes.headers.entries()) {
         proxyRes.setHeader(key, val);
     }
-    workerSystemLog('debug', 'Http proxy headers set');
     proxyRes.setHeader(invocationIdHeader, invocationId);
     proxyRes.statusCode = userRes.status;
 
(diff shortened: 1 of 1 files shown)


src/http/httpProxy.ts
```diff
@@ -105,8 +106,24 @@ export async function setupHttpProxy(): Promise<string> {
 
         server.listen(() => {
             const address = server.address();
+            // Valid address has been created
             if (address !== null && typeof address === 'object') {
-                resolve(`http://localhost:${address.port}/`);
+                if (address.port === 0) {
+                    // Auto-assigned port is 0, find and bind to an open port
+                    workerSystemLog('debug', `Port 0 assigned. Finding open port.`);
+                    findOpenPort(51929, (openPort: number) => {
```
gavin-aguiar: Use this as a constant above and set it to 55000

src/http/httpProxy.ts
```diff
@@ -117,3 +136,37 @@ export async function setupHttpProxy(): Promise<string> {
...
 }
+
+// Function to get a random port within a specified range
+function getRandomPort(): number {
+    return Math.floor(Math.random() * (maxPort - minPort + 1)) + minPort;
+}
+
+// Function to find an open port starting from a specified port
+function findOpenPort(callback: (port: number) => void): void {
+    const server = net.createServer();
```
gavin-aguiar, line 142: Why are we creating a new server? Can 't we use server already created?

src/http/httpProxy.ts
hallvictoria: It is a constant

src/http/httpProxy.ts
hallvictoria: The server we created is already listening in on port 0. We could use that one, but it would be closing and re-listening a lot. I created a new one to keep the logic cleaner and because using `net.createServer()` should be faster when just checking the ports.

src/http/httpProxy.ts
```diff
@@ -117,3 +136,37 @@ export async function setupHttpProxy(): Promise<string> {
...
+
+    function tryPort(port: number) {
+        server.once('error', () => {
```
gavin-aguiar, line 152: I think we should have a retry limit. If the vnet has blocked all ports, this will lead to stack overflow.

gavin-aguiar, line 152: This will retry finding a new port for all errors. How about we check if (err.code === 'EADDRINUSE')  and then find the next port.

src/http/httpProxy.ts
```diff
@@ -117,3 +136,37 @@ export async function setupHttpProxy(): Promise<string> {
...
+            // If the port is unavailable, get a random port and try again
+            const randomPort = getRandomPort();
+            tryPort(randomPort);
+        });
+
```
gavin-aguiar, line 156: Add `server.removeAllListeners();` . This is to remove all listeners for each recursive call. 
https://nodejs.org/api/events.html#nodeeventtargetremovealllistenerstype

src/http/httpProxy.ts
hallvictoria: We now only try 25 ports. If none are found, an error is thrown.

src/http/httpProxy.ts
hallvictoria: With the recursive strategy, this is already applied.

src/http/httpProxy.ts
hallvictoria: Discussed offline.

gavin-aguiar (approved):

PR #350, opened 2025-06-04
Updating the undici Major Version

swapnil-nagar:


hallvictoria (approved):

BridgeAR:
This breaks Node.js 18 support, due to unidici 7 only supporting Node.js 20 and above. Latest undici 6 supports >=18.17, while the latest undici release that had full Node.js 18 support is 6.13.0. Since 6.13.0 is vulnerable, I would just plainly revert this commit.

swapnil-nagar:
> This breaks Node.js 18 support, due to unidici 7 only supporting Node.js 20 and above. Latest undici supports >=18.17, while the latest undici release that had full Node.js 18 support is 6.13.0. Since 6.13.0 is vulnerable, I would just plainly revert this commit.

I really want to thank you for surfacing the issue and we have deprecated the 4.7.3 version.

timtucker-dte:
> > This breaks Node.js 18 support, due to unidici 7 only supporting Node.js 20 and above. Latest undici supports >=18.17, while the latest undici release that had full Node.js 18 support is 6.13.0. Since 6.13.0 is vulnerable, I would just plainly revert this commit.
> 
> I really want to thank you for surfacing the issue and we have deprecated the 4.7.3 version.

For clarity, given that Node 18 is no longer a supported runtime for Azure Functions why does it matter if newer versions of this library no longer support it?

https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=isolated-process%2Cv4&pivots=programming-language-javascript

July 10, 2025 is the end of support date for other Azure SDKs, so it's not unreasonable to be dropping support here as well:
https://devblogs.microsoft.com/azure-sdk/announcing-the-end-of-support-for-node-js-18-x-in-the-azure-sdk-for-javascript/

BridgeAR:
The library itself just defines a different support range. I can actually not judge much upon the impact.
I guess with Node.js 18 support being dropped, a new major would be released?

BridgeAR:
> I really want to thank you for surfacing the issue and we have deprecated the 4.7.3 version.

You're welcome, I hope I was indeed of help and it's not too much work handling the deprecated version.

PR #357, opened 2025-06-27
Revert "Updating the undici Major Version (#350)"

BridgeAR:
This reverts commit c543818f34ebc9f939c08d96d94aa9204ece1618.

The support range for undici in newer versions does not match the current @azure/functions support range.

The original commit broke Node.js 18 support, due to unidici 7 only supporting Node.js 20 and above. Latest undici 6 supports >=18.17, while the latest undici release that had full Node.js 18 support is v6.13.0. Since v6.13.0 has known security issues, it's just a plain revert.

Ref: #350

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@azure/functions",
-    "version": "4.7.3",
+    "version": "4.7.4",
     "description": "Microsoft Azure Functions NodeJS Framework",
     "keywords": [
         "azure",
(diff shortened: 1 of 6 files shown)


BridgeAR:
@microsoft-github-policy-service agree company="Datadog"

swapnil-nagar (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@azure/functions",
-    "version": "4.7.3",
+    "version": "4.7.4",
```
swapnil-nagar, line 3: This is just reverting the code we do not need to update the version, can you revert to 4.7.2.

swapnil-nagar (changes requested):

PR #366, opened 2025-08-29
Mcp Tool Trigger Support

swapnil-nagar:
Adding Support for MCP Tool Trigger


types/trigger.d.ts
```diff
@@ -90,6 +91,11 @@ export function mySql(options: MySqlTriggerOptions): MySqlTrigger;
  */
 export function webPubSub(options: WebPubSubTriggerOptions): WebPubSubTrigger;
 
+/**
+ * [Link to docs and examples](//TODO Add link to docs and examples)
```
hallvictoria, line 95: q: do we have docs on this?

hallvictoria (approved):

types/trigger.d.ts
swapnil-nagar: Yes we will add docs to this once we release it officially.

PR #371, opened 2025-10-14
SDK Binding GA Changes

swapnil-nagar:
This PR introduces support for SDK binding through _modelBindingData_ and _collectionModelBindingData_ parameters in the Azure Functions Node.js library, along with ServiceBus trigger enhancements.

**Key Changes**
Core RPC Data Conversion Enhancement
Added _modelBindingData_  support: Enhanced _fromRpcTypedData_ to handle individual model binding data objects with automatic client creation through _ResourceFactoryResolver_

Added _collectionModelBindingData_ support: Added support for collections of model binding data objects, processing the first item in the collection as the source
Error handling: Comprehensive error handling with user-friendly messages when extensions library is not registered
Resource factory integration: Integration with @azure/functions-extensions-base for automatic SDK client creation

**ServiceBus Trigger Enhancements**
Type flexibility: Updated ServiceBusQueueHandler to accept unknown | any for broader message type compatibility
Auto-complete control: Added autoCompleteMessages option for controlling automatic message completion
SDK binding flag: Added sdkBinding boolean option for enabling SDK-based operations


package.json
```diff
@@ -28,7 +28,7 @@
         "README.md"
     ],
     "engines": {
-        "node": ">=18.0"
+        "node": ">=20.0"
```
hallvictoria, line 31: This means that only Node 20 apps will be able to download the latest library version, right? We could also look at bumping the `undici` major version now (or removing that dependency all together)

package.json
```diff
@@ -41,6 +41,7 @@
         "watch": "webpack --watch --mode development"
     },
     "dependencies": {
+        "@azure/functions-extensions-base": "0.2.0-preview",
```
hallvictoria, line 44: We should release a non-preview version of the base extension for GA

package.json
swapnil-nagar: Lets discuss this in Scrum today, should we bump the version?

package.json
swapnil-nagar: Good Catch. I will create a PR there.

PR #372, opened 2025-10-14
MCP Tool trigger required parameter breaking change

swapnil-nagar:


hallvictoria (approved):

gavin-aguiar (approved):

types/mcpTool.d.ts
```diff
@@ -103,5 +103,5 @@ export interface McpToolProperty {
     /**
      * Indicates whether the property is required.
      */
-    required?: boolean;
```
larohra, line 106: Was this version released with `required` or is it a hotfix change? Basically, do we need to be cautious about the older versions?

swapnil-nagar: This package was in preview status for a while and never GA'd. This change is breaking change but wont impact the customer as this was not released officially.

PR #377, opened 2025-10-30
MCP ToolTrigger Fulent Property and IsArray support

swapnil-nagar:
Mcp changes to support following contract and it is backward compatible.

```
import { app, InvocationContext, output, arg, input } from "@azure/functions";

const PROPERTY_TYPE = "string";

// Hello function - responds with hello message
export async function mcpToolHello(_toolArguments:unknown, context: InvocationContext): Promise<string> {
    console.log(_toolArguments);
    // Get name from the tool arguments
    const mcptoolargs = context.triggerMetadata.mcptoolargs as {
        name?: string;
    };
    const name = mcptoolargs?.name;

    console.info(`Hello ${name}, I am MCP Tool!`);

    return `Hello ${name}, I am MCP Tool!`;
}

// Register the hello tool
app.mcpTool('hello', {
    toolName: 'hello',
    description: 'Simple hello world MCP Tool that responses with a hello message.',
    toolProperties:{
        name: arg.string().describe('Required property to identify the caller.').optional(),
        arrayT: arg.string().asArray().describe('An array of strings property.'),
        ats: arg.object().asArray().describe('An array of objects property.'),
        ats1: arg.number().describe('A number property.'),
        ats2: arg.boolean().describe('A boolean property.'),
        ats3: arg.integer().describe('An integer property.'),
        ats4: arg.datetime().describe('A dateTime property.')
    },
    handler: mcpToolHello
});
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@azure/functions",
-    "version": "4.8.0",
+    "version": "4.8.1",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@azure/functions",
-            "version": "4.8.0",
+            "version": "4.8.1",
             "license": "MIT",
             "dependencies": {
                 "cookie": "^0.7.0",
@@ -66,9 +66,9 @@
             }
         },
         "node_modules/@babel/helper-validator-identifier": {
-            "version": "7.27.1",
-            "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.27.1.tgz",
-            "integrity": "sha512-D2hP9eA+Sqx1kBZgzxZh0y1trbuU+JoDkiEwqhQ36nodYqJwyEIhPSdMNd7lOm/4io72luTPWH20Yda0xOuUow==",
+            "version": "7.28.5",
+            "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.28.5.tgz",
+            "integrity": "sha512-qSs4ifwzKJSV39ucNjsvc6WVHs6b7S03sOh2OcHF9UHfVPqWWALUsNUVzhSBiItjRZoLHx7nIarVjqKVusUZ1Q==",
             "dev": true,
             "license": "MIT",
             "engines": {
@@ -180,9 +180,9 @@
             }
         },
         "node_modules/@eslint-community/eslint-utils": {
-            "version": "4.7.0",
-            "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.7.0.tgz",
-            "integrity": "sha512-dyybb3AcajC7uha6CvhdVRJqaKyn7w2YKqKyAN37NKYgZT36w+iRb0Dymmc5qEJ549c/S31cMMSFd75bteCpCw==",
+            "version": "4.9.0",
+            "resolved": "https://registry.npmjs.org/@eslint-community/eslint-utils/-/eslint-utils-4.9.0.tgz",
+            "integrity": "sha512-ayVFHdtZ+hsq1t2Dy24wCmGXGe4q9Gu3smhLYALJrr473ZH27MsnSL+LKUlimp4BWJqMDMLmPpx/Q9R3OAlL4g==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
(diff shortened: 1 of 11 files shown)


types/mcpTool.d.ts
```diff
@@ -48,9 +48,9 @@ export interface McpToolTriggerOptions {
 
     /**
      * Additional properties or metadata for the tool.
-     * This is a dictionary of key-value pairs that can be used to configure the trigger.
+     * Can be provided as an array (legacy format) or as a toolProp object format.
      */
-    toolProperties?: any | McpToolProperty[];
+    toolProperties?: McpToolProperty[] | ToolProps | any;
```
swapnil-nagar: Remove any make it stricter.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "@azure/functions",
-    "version": "4.8.0",
+    "version": "4.8.1",
```
hallvictoria, line 3: Checking, should `package.json` / `constants.ts` be changed as well?

src/index.ts
```diff
@@ -16,6 +16,7 @@ export { InvocationContext } from './InvocationContext';
 export * as output from './output';
 export * as trigger from './trigger';
 export { Disposable } from './utils/Disposable';
+export { toolProp } from './utils/toolProperties';
```
hallvictoria: Is `toolProp` the name exposed to the customer? Just wondering if we should call it `toolProperty` instead, but open to discussion

src/utils/toolProperties.ts
```diff
@@ -0,0 +1,252 @@
...
+    object(): ToolPropertyBuilder {
+        this.property.propertyType = 'object';
+        return this;
+    }
+
+    /**
+     * Set the property type to long
+     */
+    long(): ToolPropertyBuilder {
+        this.property.propertyType = 'long';
```
hallvictoria: Do you also need to add `double()`? (unsure about all the supported types)

src/utils/toolProperties.ts
```diff
@@ -0,0 +1,252 @@
...
+        return this;
+    }
+
+    /**
+     * Set the description for the property
+     * @param description - Description of the property's purpose
+     */
+    desc(description: string): ToolPropertyBuilder {
+        this.property.description = description;
```
hallvictoria, line 90: Wondering if this could be renamed to `description` instead? (I initially thought `desc` was transforming the input to lowercase)

src/utils/toolProperties.ts
```diff
@@ -0,0 +1,252 @@
...
+     * @private
+     */
+    private buildInternal(): McpToolProperty {
+        if (!this.property.propertyType) {
+            throw new Error('Property type must be specified (use .string(), .number(), etc.)');
+        }
+
+        return {
+            propertyName: '', // Will be set by the consumer based on the key
+            propertyType: this.property.propertyType,
```
hallvictoria, line 121: Should these use the `get` methods defined above?

src/utils/toolProperties.ts
```diff
@@ -0,0 +1,252 @@
...
+ * @param toolProps - Object with property names as keys and ToolProperty as values
+ * @returns Array of McpToolProperty objects with propertyName set correctly
+ */
+export function convertToolProperties(toolProps: ToolProps): McpToolProperty[] {
+    return Object.entries(toolProps).map(([propertyName, property]) => ({
+        propertyName,
+        propertyType: property.propertyType,
+        description: property.description || '', // Default to empty string if not provided
+        isRequired: property.isRequired,
+        isArray: property.isArray,
```
hallvictoria, line 201: Should `propertyName`, `isArray`, and `isRequired` also have defaults?

package-lock.json
swapnil-nagar: That has been changed with the last PR, we were waiting on releasing this final changes.

src/index.ts
swapnil-nagar: Thats a valid point but that was too long and 'property' was too generic without giving the details. hence went with toolProp but open to change.

src/utils/toolProperties.ts
swapnil-nagar: I am asking for list from Fabio, I will add double for now.

src/utils/toolProperties.ts
swapnil-nagar: Valid point, people familiar with zod, uses  describe, I will change to describe.

src/utils/toolProperties.ts
swapnil-nagar: propertyName is required property,

isArray and isRequired defaulted at line 33 and 29 respectively.

src/utils/toolProperties.ts
swapnil-nagar: During the toolPropertyBuilder these are set already so we should be good.

src/index.ts
vrdmr: Can it be `toolArg` or `Arg` only?

swapnil-nagar: changed to arg

src/utils/toolProperties.ts
swapnil-nagar: double is number according to Fabio hence I have not supporting that type.

Below are the list of types that are supported.

MCP, the following type conversions are automatically handled by the worker:
string, DateTime, DateTimeOffset, Guid, char -> string 
int -> integer
bool -> boolean
numeric types (outside of int) -> number

hallvictoria (approved):

PR #394, opened 2025-11-21
Upgrading to Typescript 5

swapnil-nagar:


TsuyoshiUshio (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@azure/functions",
-    "version": "4.9.0",
+    "version": "4.9.1",
```
gavin-aguiar: This is a big change. We should consider updating the major version.  4.9.1 is only a patch version update.

package.json
```diff
@@ -79,9 +79,9 @@
         "prettier": "^2.4.1",
         "semver": "^7.3.5",
         "sinon": "^20.0.0",
-        "ts-loader": "^9.3.1",
-        "ts-node": "^3.3.0",
-        "typescript": "^4.9.5",
+        "ts-loader": "^9.5.1",
+        "ts-node": "^10.9.2",
+        "typescript": "~5.0.0",
```
gavin-aguiar, line 84: Are there any breaking changes? have we run any tests with it?

swapnil-nagar: No there is no breaking change. The existing app will be using the typescript 5 and E2E test can verify that.

package.json
swapnil-nagar: Sure, I can bump the major version for it.

package.json
gavin-aguiar: Breaking changes would be for the customer. We might not have an exhaustive test coverage. 
https://www.webdevtutor.net/blog/typescript-breaking-changes-4-to-5

gavin-aguiar (approved):

package.json
swapnil-nagar: We discussed this offline, since the trans piled code gets available for customer this change should be fine and we validate by testing apps.

PR #395, opened 2025-11-25
Remove Undici Dependency

swapnil-nagar:
Description:

Removed undici dependency completely
Replaced all undici usage with native Node.js fetch API (Request, Response, Headers, FormData, File).

Reason:

Headers is now a global type from Node.js's built-in fetch API (available in Node.js 18+). When we removed undici, we didn't need to change the Headers type reference because Node.js provides the same Headers class natively - it's automatically available in the global scope just like Request, Response, and FormData. The type definition comes from @types/node@^20.0.0 which includes all the native fetch API types, so TypeScript recognizes it without any imports.

### DIFF
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -8,7 +8,6 @@ import { Blob } from 'buffer';
 import { IncomingMessage } from 'http';
 import * as stream from 'stream';
 import { ReadableStream } from 'stream/web';
-import { FormData, Headers, HeadersInit, Request as uRequest } from 'undici';
 import { URLSearchParams } from 'url';
 import { fromNullableMapping } from '../converters/fromRpcNullable';
 import { fromRpcTypedData } from '../converters/fromRpcTypedData';
@@ -17,7 +16,7 @@ import { isDefined, nonNullProp } from '../utils/nonNull';
 import { extractHttpUserFromHeaders } from './extractHttpUserFromHeaders';
 
 interface InternalHttpRequestInit extends RpcHttpData {
-    undiciRequest?: uRequest;
+    nativeRequest?: Request;
 }
 
 export class HttpRequest implements types.HttpRequest {
@@ -25,14 +24,14 @@ export class HttpRequest implements types.HttpRequest {
     readonly params: HttpRequestParams;
 
     #cachedUser?: HttpRequestUser | null;
-    #uReq: uRequest;
+    #nativeReq: Request;
     #init: InternalHttpRequestInit;
 
     constructor(init: InternalHttpRequestInit) {
         this.#init = init;
 
-        let uReq = init.undiciRequest;
-        if (!uReq) {
+        let nativeReq = init.nativeRequest;
+        if (!nativeReq) {
             const url = nonNullProp(init, 'url');
 
             let body: Buffer | string | undefined;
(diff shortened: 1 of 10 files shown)


src/http/extractHttpUserFromHeaders.ts
gavin-aguiar, line 10: If we have removed the import, where is Headers coming from in this line?

src/http/HttpRequest.ts
```diff
@@ -8,7 +8,6 @@ import { Blob } from 'buffer';
 import { IncomingMessage } from 'http';
 import * as stream from 'stream';
 import { ReadableStream } from 'stream/web';
-import { FormData, Headers, HeadersInit, Request as uRequest } from 'undici';
```
gavin-aguiar, line 11: Same question here, all these imports are used in this file. It will fail right if these are removed?

src/http/extractHttpUserFromHeaders.ts
swapnil-nagar: So Headers is now a global type from Node.js's built-in fetch API and we do not support Node 18 or earlier versions of node.js that change should be Ok.

src/http/HttpRequest.ts
swapnil-nagar: When we removed undici, we didn't need to change the Headers type reference because Node.js provides the same Headers class natively. (Starting Node 18+)

gavin-aguiar: Same with Formdata?

gavin-aguiar (approved):

src/http/HttpRequest.ts
swapnil-nagar: Yes same with formData available with the typescript library.

pragmatta:
Removing BodyInit, HeadersInit from http.d.ts is causing my Azure Function build break with tsc. Downgrading to 4.10.0 helps.

```
[func] node_modules/@azure/functions/types/http.d.ts(206,12): error TS2304: Cannot find name 'BodyInit'.
[func] node_modules/@azure/functions/types/http.d.ts(223,15): error TS2304: Cannot find name 'HeadersInit'.
```

swapnil-nagar:
> Removing BodyInit, HeadersInit from http.d.ts is causing my Azure Function build break with tsc. Downgrading to 4.10.0 helps.
> 
> 
> 
> ```
> 
> [func] node_modules/@azure/functions/types/http.d.ts(206,12): error TS2304: Cannot find name 'BodyInit'.
> 
> [func] node_modules/@azure/functions/types/http.d.ts(223,15): error TS2304: Cannot find name 'HeadersInit'.
> 
> ``` 

@pragmatta What version on Node you have?

pragmatta:
Node v22.16.0, tsc 5.9.2.

swapnil-nagar:
> Node v22.16.0, tsc 5.9.2.

@pragmatta I just validated few scenarios at our end and http scenario works fine for me. 

Is this issue is happening on dev machine or on the cloud?

pragmatta:
Both, it's just running tsc for normal build. Maybe if your tsconfig has something that introduces those but they are not found in that file.

```
tsc
node_modules/@azure/functions/types/http.d.ts:206:12 - error TS2304: Cannot find name 'BodyInit'.

206     body?: BodyInit;
               ~~~~~~~~

node_modules/@azure/functions/types/http.d.ts:223:15 - error TS2304: Cannot find name 'HeadersInit'.

223     headers?: HeadersInit;
                  ~~~~~~~~~~~


Found 2 errors in the same file, starting at: node_modules/@azure/functions/types/http.d.ts:206
```

swapnil-nagar:
Can you provide your tsconfig.json,  I would like to try at our end and check if we can reproduce.

pragmatta:
Tried upgrading module/moduleResolution to nodenext and lib to ES2022/23/34 but did not change this.
```javascript
{
  "compilerOptions": {
    "module": "node16",
    "moduleResolution": "node16",
    "target": "ES2022",
    "lib": ["ES2021"],
    "outDir": "dist",
    "rootDir": ".",
    "sourceMap": true,
    "strict": true
  }
 }
```

swapnil-nagar:
> Tried upgrading module/moduleResolution to nodenext and lib to ES2022/23/34 but did not change this.
> 
> ```js
> {
>   "compilerOptions": {
>     "module": "node16",
>     "moduleResolution": "node16",
>     "target": "ES2022",
>     "lib": ["ES2021"],
>     "outDir": "dist",
>     "rootDir": ".",
>     "sourceMap": true,
>     "strict": true
>   }
>  }
> ```

Ok, since you have defined the lib explicitly, you need to define DOM as well or remove the lib from the tsconfig altogether.

Option 1:

    "lib": ["ES2022", "DOM"]
    
Option 2:

Remove lib

pragmatta:
Yes, you are correct! Both of those work, thank you. I have to think a bit if it matters which way I go.

Issue #397, opened 2025-12-05
Bug: library .d.ts files should not import its .ts files - this can cause compilation errors in user projects


azerum:
# Overivew

When library .d.ts files import .ts files, those .ts files are included in the compilation step of the user projects. For example, since `index.d.ts` [imports toolProperties.ts here](https://github.com/Azure/azure-functions-nodejs-library/blob/70bcdf41e8f7319c0ec16a84472d6fd38582fcfc/types/index.d.ts#L34), `toolProperties.ts` will be compiled as part of the user project

This is problematic since file may not compile under `tsconfig.json` settings of the user

For example, this library is targets CJS, but uses ESM syntax (export). If users use `verbatimModuleSyntax: true` - which requires that CJS files use CJS syntax - **their compilation will fail, with no easy fix**

Similarly, if users use any `strict*` option this project doesn't, `toolProperties.ts` may not compile

Effectively, `toolProperties.ts` of this library limits what `tsconfig.json` settings users can use

# Expected behavior

Library's `.d.ts` files may import only other `.d.ts` files, but not `.ts` files

# Actual behavior

`index.d.ts` [imports toolProperties.ts here](https://github.com/Azure/azure-functions-nodejs-library/blob/70bcdf41e8f7319c0ec16a84472d6fd38582fcfc/types/index.d.ts#L34)

Example user project that fails to compile due to this: https://github.com/azerum/azure-functions-type-check-bug

# Suggested fix

Generate `.d.ts` file for `toolProperties.ts` and export that. I am not familiar with why `.d.ts` files of this library are written manually and not generated from `.ts` files, I assume there must be a good reason..

swapnil-nagar:
This is fixed as part for following PR
https://github.com/Azure/azure-functions-nodejs-library/pull/399

PR #399, opened 2026-01-02
Fixing the verbatimModuleSyntax issue

swapnil-nagar:
Overivew
When library .d.ts files import .ts files, those .ts files are included in the compilation step of the user projects. For example, since index.d.ts [imports toolProperties.ts here](https://github.com/Azure/azure-functions-nodejs-library/blob/70bcdf41e8f7319c0ec16a84472d6fd38582fcfc/types/index.d.ts#L34), toolProperties.ts will be compiles as part of the user project

This is problematic since file may not compile under tsconfig.json settings of the user

For example, this library is targets CJS, but uses ESM syntax (export). If users use verbatimModuleSyntax: true - which requires that CJS files use CJS syntax - their compilation will fail, with no easy fix

Similarly, if users if users use any strict* option this project doesn't, toolProperties.ts may not compile

Effectively, toolProperties.ts of this library limits what tsconfig.json settings users can use

Expected behavior
Library's .d.ts files may import only other .d.ts files, but not .ts files

Actual behavior
index.d.ts [imports toolProperties.ts here](https://github.com/Azure/azure-functions-nodejs-library/blob/70bcdf41e8f7319c0ec16a84472d6fd38582fcfc/types/index.d.ts#L34)

Example user project that fails to compile due to this: https://github.com/azerum/azure-functions-type-check-bug

Suggested fix
Generate .d.ts file for toolProperties.ts and export that. I am not familiar with why .d.ts files of this library are written manually and not generated from .ts files, I assume there must be a good reason..

### DIFF
--- a/types/toolProperties.d.ts
+++ b/types/toolProperties.d.ts
@@ -1,0 +1,76 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import type { Args, McpToolProperty } from './mcpTool';
+import type { IToolPropertyBuilder } from './toolPropertyBuilder';
+
+/**
+ * Fluent API builder for creating MCP Tool properties
+ * Implements IToolPropertyBuilder interface to ensure consistency
+ */
+export declare class ToolPropertyBuilder implements IToolPropertyBuilder {
+    get propertyType(): string;
+    get description(): string;
+    get isRequired(): boolean;
+    get isArray(): boolean;
+
+    string(): IToolPropertyBuilder;
+    double(): IToolPropertyBuilder;
+    long(): IToolPropertyBuilder;
+    number(): IToolPropertyBuilder;
+    boolean(): IToolPropertyBuilder;
+    object(): IToolPropertyBuilder;
+    integer(): IToolPropertyBuilder;
+    datetime(): IToolPropertyBuilder;
+    describe(description: string): IToolPropertyBuilder;
+    optional(): McpToolProperty;
+    asArray(): IToolPropertyBuilder;
+}
+
+/**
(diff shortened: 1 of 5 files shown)


hallvictoria (dismissed):

types/toolProperties.d.ts
```diff
@@ -0,0 +1,76 @@
...
+    string(): IToolPropertyBuilder;
+    double(): IToolPropertyBuilder;
+    long(): IToolPropertyBuilder;
+    number(): IToolPropertyBuilder;
+    boolean(): IToolPropertyBuilder;
+    object(): IToolPropertyBuilder;
+    integer(): IToolPropertyBuilder;
+    datetime(): IToolPropertyBuilder;
+    describe(description: string): IToolPropertyBuilder;
+    optional(): McpToolProperty;
```
TsuyoshiUshio: optional() type is `Arg` for `IToolPropertyBuilder` interface. It works, however, might better to use the same type.

TsuyoshiUshio:
Hi @swapnil-nagar  The change totally makes sense. Leave small nit. but I'd happy to approve. 
One question is, do you have issue that related this PR? If so, let's link it.

VanTanev:
> Hi @swapnil-nagar The change totally makes sense. Leave small nit. but I'd happy to approve. One question is, do you have issue that related this PR? If so, let's link it.

Not op, but an issue exists: #397

types/toolProperties.d.ts
swapnil-nagar: Good point, addressed your comment.

TsuyoshiUshio (approved):
LGTM

PR #401, opened 2026-01-21
Removing Long from dependencies

swapnil-nagar:
We use long only as a devDependency and do not rely on it at runtime, so we’re moving it to devDependencies.

We will validate whether this change has any impact on existing apps during integration testing. If a customer’s app depends on long, they will need to explicitly declare it in their own package.json.


TsuyoshiUshio (approved):
LGTM.

Issue #404, opened 2026-01-26
Build fails after removing undici dependency in v4.11.0


DanielSed:
The [removal of the undici dependency ](https://github.com/Azure/azure-functions-nodejs-library/pull/395) is causing our Azure Function build to fail as the `BodyInit` and `HeaderInit` type definitions are still being referenced in [types/http.d.ts](https://github.com/Azure/azure-functions-nodejs-library/blob/0288d8b59891abb61d6ac73038bda4ca37f726ab/types/http.d.ts).

We are getting the following error:
```
node_modules/@azure/functions/types/http.d.ts(206,12): error TS2304: Cannot find name 'BodyInit'.
node_modules/@azure/functions/types/http.d.ts(223,15): error TS2304: Cannot find name 'HeadersInit'.
```

The issue is discussed in the [Remove Undici Dependency](https://github.com/Azure/azure-functions-nodejs-library/pull/395) pull request. The provided [solution](https://github.com/Azure/azure-functions-nodejs-library/pull/395#issuecomment-3781528178) was to add the DOM library to lib in tsconfig.json (our lib is set to ES2018). We want to avoid that as that would clutter our app with unneeded browser-specific APIs.

Please fix this.

swapnil-nagar:
Understandable that you do not want to have DOM lib as part of your function app. We will prioritize a fix and share the details soon.

swapnil-nagar:
Just release the new version of library with fix, let us know if you still experience issue.
https://www.npmjs.com/package/@azure/functions/v/4.11.1

PR #405, opened 2026-01-28
Removing lib.DOM dependency

swapnil-nagar:
Removing the lib.DOM Dependency

Issue: https://github.com/Azure/azure-functions-nodejs-library/issues/404

### DIFF
--- a/types/http.d.ts
+++ b/types/http.d.ts
@@ -7,6 +7,20 @@ import { URLSearchParams } from 'url';
 import { FunctionOptions, FunctionOutput, FunctionResult, FunctionTrigger } from './index';
 import { InvocationContext } from './InvocationContext';
 
+/**
+ * Represents the body types that can be used in an HTTP response.
+ * This is a local definition to avoid dependency on lib.dom.
+ * Compatible with Node.js native Fetch API body types.
+ */
+export type HttpResponseBodyInit = ReadableStream | Blob | ArrayBufferView | ArrayBuffer | string | null | undefined;
+
+/**
+ * Represents the headers types that can be used to initialize HTTP headers.
+ * This is a local definition to avoid dependency on lib.dom.
+ * Compatible with Node.js native Fetch API header types.
+ */
+export type HttpHeadersInit = Headers | Record<string, string> | [string, string][];
+
 export type HttpHandler = (
     request: HttpRequest,
     context: InvocationContext
@@ -203,7 +217,7 @@ export interface HttpResponseInit {
     /**
      * HTTP response body
      */
-    body?: BodyInit;
+    body?: HttpResponseBodyInit;
 
     /**
      * A JSON-serializable HTTP Response body.
@@ -220,7 +234,7 @@ export interface HttpResponseInit {
     /**
      * HTTP response headers
      */
-    headers?: HeadersInit;
+    headers?: HttpHeadersInit;
 
     /**
      * HTTP response cookies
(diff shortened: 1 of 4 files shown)


types/http.d.ts
```diff
@@ -7,6 +7,20 @@ import { URLSearchParams } from 'url';
 import { FunctionOptions, FunctionOutput, FunctionResult, FunctionTrigger } from './index';
 import { InvocationContext } from './InvocationContext';
 
+/**
+ * Represents the body types that can be used in an HTTP response.
+ * This is a local definition to avoid dependency on lib.dom.
+ * Compatible with Node.js native Fetch API body types.
+ */
+export type HttpResponseBodyInit = ReadableStream | Blob | ArrayBufferView | ArrayBuffer | string | null | undefined;
```
TsuyoshiUshio: ## Potential Issues with Type Definitions

Thank you for this PR! Removing the `lib.DOM` dependency is the right approach. However, I noticed a few potential compatibility gaps in the new type definitions:

### 1. Missing `FormData` and `URLSearchParams` in `HttpResponseBodyInit`

The standard `BodyInit` type (from Web/Node.js Fetch API) includes:
```typescript
type BodyInit = ReadableStream | Blob | ArrayBufferView | ArrayBuffer | FormData | URLSearchParams | string
```

The PR's definition:
```typescript
type HttpResponseBodyInit = ReadableStream | Blob | ArrayBufferView | ArrayBuffer | string | null | undefined
```

**Issue**: `FormData` and `URLSearchParams` are missing. Users who pass these types as the response body will encounter type errors, even though it works at runtime.

### 2. `HttpHeadersInit` Missing `Iterable` Support

The standard `HeadersInit` type:
```typescript
type HeadersInit = Headers | string[][] | Record<string, string> | Iterable<[string, string]>
```

The PR's definition:
```typescript
type HttpHeadersInit = Headers | Record<string, string> | [string, string][]
```

**Issue**: `Iterable<[string, string]>` is not supported, so passing a `Map<string, string>` or other custom iterables will fail type checking.

### 3. Type Cast in `HttpResponse.ts`

```typescript
this.#nativeRes = new Response(init.body as BodyInit, resInit);
```

This cast bypasses the type system. If `HttpResponseBodyInit` is narrower than what Node.js's `Response` actually accepts, there could be inconsistencies where valid runtime values fail type checking.

### Suggested Fix

```typescript
export type HttpResponseBodyInit = 
    | ReadableStream 
    | Blob 
    | ArrayBufferView 
    | ArrayBuffer 
    | FormData 
    | URLSearchParams 
    | string 
    | null 
    | undefined;

export type HttpHeadersInit = 
    | Headers 
    | Record<string, string> 
    | Iterable<[string, string]>;
```

> **Note**: `FormData` and `URLSearchParams` are globally available in Node.js 18+, which aligns with the minimum supported version for Azure Functions Node.js v4 programming model.

### 4. Test Coverage

Consider adding tests for `FormData` and `URLSearchParams` body types to ensure full coverage.

---

Overall, this is a great improvement! Just these minor additions would ensure full backward compatibility with the original `BodyInit` and `HeadersInit` types.

swapnil-nagar: I have added the tests related to FormData and URLSearchParams.

swapnil-nagar: Note: Map<string, string> and other iterables are supported since they satisfy the [string, string][] array interface, so no additional change is needed.

TsuyoshiUshio (approved):

PR #407, opened 2026-01-30
Fixing the Cloning issue with the library

swapnil-nagar:
If you use `clone` method on request object in HTTP triggered function with POST method, an exception is thrown.

Happens in version 4.11.0, if you roll back to 4.10.0 the issue is gone.

## Stack trace:
```
Result: Failure Type: Exception: Cannot clone object of unsupported type. Stack: DataCloneError: Cannot clone object of unsupported type. at new DOMException (node:internal/per_context/domexception:76:18) at structuredClone (node:internal/worker/js_transferable:126:26) at HttpRequest.clone (/home/site/wwwroot/node_modules/@azure/functions/dist/azure-functions.js:1878:25) at /home/site/wwwroot/dist/src/functions/HandleSlackAchievementCommand.js:26:82 at Generator.next (<anonymous>) at /home/site/wwwroot/dist/src/functions/HandleSlackAchievementCommand.js:8:71 at new Promise (<anonymous>) at __awaiter (/home/site/wwwroot/dist/src/functions/HandleSlackAchievementCommand.js:4:12) at handleSlackAchievementCommand (/home/site/wwwroot/dist/src/functions/HandleSlackAchievementCommand.js:24:12) at InvocationModel.<anonymous> (/home/site/wwwroot/node_modules/@azure/functions/dist/azure-functions.js:215:46)
```

## Environment
- Node 24
- Cloud deployment, Linux
- Flex Plan
- @azure/functions = 4.11.0

### DIFF
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -109,7 +109,9 @@ export class HttpRequest implements types.HttpRequest {
     }
 
     clone(): HttpRequest {
-        const newInit = structuredClone(this.#init);
+        // Exclude nativeRequest from structuredClone since Request objects can't be cloned that way
+        const { nativeRequest: _nativeRequest, ...initWithoutNativeReq } = this.#init;
+        const newInit: InternalHttpRequestInit = structuredClone(initWithoutNativeReq);
         newInit.nativeRequest = this.#nativeReq.clone();
         return new HttpRequest(newInit);
     }
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -79,7 +79,7 @@ export class HttpRequest implements types.HttpRequest {
     }
 
     get body(): ReadableStream<any> | null {
-        return this.#nativeReq.body as ReadableStream<any> | null;
+        return this.#nativeReq.body;
     }
 
     get bodyUsed(): boolean {
@@ -91,10 +91,12 @@ export class HttpRequest implements types.HttpRequest {
     }
 
     async blob(): Promise<Blob> {
-        return this.#nativeReq.blob() as Promise<Blob>;
+        return this.#nativeReq.blob();
     }
 
+    // eslint-disable-next-line deprecation/deprecation
     async formData(): Promise<FormData> {
+        // eslint-disable-next-line deprecation/deprecation
         return this.#nativeReq.formData();
     }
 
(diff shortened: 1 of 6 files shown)


src/http/HttpRequest.ts
TsuyoshiUshio, line 111: Thanks for the PR and the additional test coverage! 

I have a concern about whether this fully addresses the root cause. The error occurs in the `clone()` method when `structuredClone()` tries to clone a `Request` object, which is not supported.

### The Issue

Looking at the current `clone()` implementation:

```typescript
clone(): HttpRequest {
    const newInit = structuredClone(this.#init);  // ← Problem: #init may contain nativeRequest
    newInit.nativeRequest = this.#nativeReq.clone();
    return new HttpRequest(newInit);
}
```

When `HttpRequest` is created via `createStreamRequest()` (which is used by the Azure Functions runtime for HTTP streaming), the `#init` object contains a `nativeRequest` property:

```typescript
// From createStreamRequest()
return new HttpRequest({
    nativeRequest: nativeReq,  // ← This gets stored in #init
    params,
});
```

`structuredClone()` cannot clone `Request` objects, causing `DataCloneError`.

### Suggested Test Case

The current tests create `HttpRequest` without `nativeRequest` in the init object, so they don't reproduce the issue. This test should fail with the current implementation:

```typescript
it('clone with nativeRequest in init (simulating createStreamRequest path)', () => {
    // This simulates how createStreamRequest() creates HttpRequest
    // with nativeRequest already in the init object
    const nativeReq = new Request('http://localhost:7071/api/test', {
        method: 'POST',
        body: 'test body',
    });
    
    const req = new HttpRequest({
        nativeRequest: nativeReq,
        params: { id: '123' },
    } as any);
    
    // This should throw DataCloneError: Cannot clone object of unsupported type
    expect(() => req.clone()).to.not.throw();
});
```

### Suggested Fix

Exclude `nativeRequest` before calling `structuredClone()`:

```typescript
clone(): HttpRequest {
    const { nativeRequest, ...initWithoutNative } = this.#init;
    const newInit: InternalHttpRequestInit = structuredClone(initWithoutNative);
    newInit.nativeRequest = this.#nativeReq.clone();
    return new HttpRequest(newInit);
}
```

Could you verify if this test case reproduces the issue and consider adding this fix?

TsuyoshiUshio (commented):
Have a look the comment!

TsuyoshiUshio (approved):

PR #408, opened 2026-02-03
Add MCP Resource Trigger Support

swapnil-nagar:
### Summary

This PR adds support for MCP (Model Context Protocol) Resource Triggers in the Azure Functions Node.js library, enabling developers to expose resources via the MCP protocol.

## Changes
New Features
[app.mcpResource()] - New function registration method for MCP Resource triggers
[trigger.mcpResource()] - New trigger factory function for creating MCP Resource triggers
Type definitions for McpResourceTriggerOptions, McpResourceFunctionOptions, McpResourceTrigger, and McpResourceTriggerHandler

### DIFF
--- a/src/converters/toMcpResourceTriggerOptionsToRpc.ts
+++ b/src/converters/toMcpResourceTriggerOptionsToRpc.ts
@@ -1,0 +1,56 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import { McpResourceTriggerOptions, McpResourceTriggerOptionsToRpc } from '../../types';
+
+/**
+ * Converts an McpResourceTriggerOptions object to an McpResourceTriggerOptionsToRpc object.
+ *
+ * @param options - The input options to be converted.
+ * @returns The converted McpResourceTriggerOptionsToRpc object.
+ * @throws Error if required properties are missing or invalid.
+ */
+export function convertToMcpResourceTriggerOptionsToRpc(
+    options: McpResourceTriggerOptions
+): McpResourceTriggerOptionsToRpc {
+    // Validate required properties
+    if (!options.uri || typeof options.uri !== 'string' || options.uri.trim() === '') {
+        throw new Error('MCP Resource trigger requires a valid "uri" property.');
+    }
+
+    if (!options.resourceName || typeof options.resourceName !== 'string' || options.resourceName.trim() === '') {
+        throw new Error('MCP Resource trigger requires a valid "resourceName" property.');
+    }
+
+    // Build the result object with required properties
+    const result: McpResourceTriggerOptionsToRpc = {
+        uri: options.uri,
+        resourceName: options.resourceName,
+    };
+
+    // Add optional properties if they are defined
+    if (options.title !== undefined) {
+        result.title = options.title;
+    }
+
+    if (options.description !== undefined) {
+        result.description = options.description;
+    }
+
+    if (options.mimeType !== undefined) {
+        result.mimeType = options.mimeType;
+    }
+
+    if (options.size !== undefined) {
+        if (typeof options.size !== 'number' || options.size < 0) {
+            throw new Error('MCP Resource trigger "size" must be a non-negative number.');
+        }
+        result.size = options.size;
+    }
+
+    if (options.metadata !== undefined) {
+        result.metadata = options.metadata;
+    }
+
+    return result;
+}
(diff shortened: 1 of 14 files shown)


types/trigger.d.ts
```diff
@@ -96,6 +97,12 @@ export function webPubSub(options: WebPubSubTriggerOptions): WebPubSubTrigger;
  */
 export function mcpTool(options: McpToolFunctionOptions): McpToolTrigger;
 
+/**
+ * Creates an MCP Resource trigger for defining resources that can be read by MCP clients.
+ * [Link to docs and examples](//TODO Add link to docs and examples)
+ */
+export function mcpResource(options: McpResourceFunctionOptions): McpResourceTrigger;
```
liliankasem, line 104: An issue pointed out by copilot is that this should be using `McpResourceTriggerOptions` not `McpResourceFunctionOptions`. 

> McpResourceFunctionOptions extends McpResourceTriggerOptions and adds a required handler field. Users calling trigger.mcpResource() directly would be forced to provide a handler, which makes no sense for a trigger factory — the handler belongs on app.mcpResource().

This is a pre-existing pattern carried from trigger.mcpTool(), which has the same issue

src/converters/toMcpResourceTriggerOptionsToRpc.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: Another copilot item:

> McpResourceTriggerOptionsToRpc is redundant. In types/mcpResource.d.ts, McpResourceTriggerOptionsToRpc has identical fields to McpResourceTriggerOptions. For the tool trigger, the ToRpc interface serves a purpose — it converts toolProperties: McpToolProperty[] | Args to toolProperties: string. But for the resource trigger, no such conversion exists. The converter function just passes properties through. This adds unnecessary indirection.

src/converters/toMcpToolTriggerOptionsToRpc.ts
```diff
@@ -39,12 +39,27 @@ export function converToMcpToolTriggerOptionsToRpc(
...
+    }
+
     // If we successfully normalized the properties, use them
     if (normalizedProperties !== undefined) {
-        return {
+        const result: McpToolTriggerOptionsToRpc = {
             ...baseResult,
             toolProperties: JSON.stringify(normalizedProperties),
         };
+        if (mcpToolTriggerOptions.metadata !== undefined) {
```
liliankasem: You should be able to set metadata outside of `normalizedProperties` - or in better words, metadata is independent of toolProperties normalization. Attaching it inside the normalizedProperties success branch couples two unrelated concerns. If someone later added a code path that returns without normalizedProperties metadata would be silently dropped.


Metadata should be set on baseResult since it's independent of toolProperties. This avoids coupling it to the toolProperties normalization outcome.

```ts
const baseResult = {
    toolName: mcpToolTriggerOptions.toolName,
    description: mcpToolTriggerOptions.description,
    ...(mcpToolTriggerOptions.metadata !== undefined && { metadata: mcpToolTriggerOptions.metadata }),
};
```

src/converters/toMcpResourceTriggerOptionsToRpc.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (options.size !== undefined) {
+        if (typeof options.size !== 'number' || options.size < 0) {
+            throw new Error('MCP Resource trigger "size" must be a non-negative number.');
+        }
+        result.size = options.size;
+    }
+
+    if (options.metadata !== undefined) {
+        // Validate that metadata is a valid JSON string
+        if (typeof options.metadata === 'string' && options.metadata.trim() !== '') {
```
liliankasem: Both converters (toMcpResourceTriggerOptionsToRpc.ts and toMcpToolTriggerOptionsToRpc.ts) skip JSON validation for empty strings and whitespace-only strings. This will lead to an exception in the host. I would suggest tightening the validation here as only json objects are accepted. 

Whilst harmless since the host handles this, it's still wasteful to send "   " over the wire when the Node side could just skip it.

liliankasem (commented):
Mostly lgtm, have some comments. But otherwise, the binding type and properties align with the host extension contract, and the metadata addition to McpToolTriggerOptionsToRpc also correctly matches the host extension's McpToolTriggerAttribute.Metadata property.

src/converters/toMcpResourceTriggerOptionsToRpc.ts
swapnil-nagar: Json vaidation is at line 55 and line 56 of toMcpToolTriggerOptionsToRpc.ts file. If parsing fail the exception will be caught and correct error message will be thrown.

Wil that be sufficient or you want to have json validation on some other attribute/parameter.

liliankasem (dismissed):

hallvictoria (dismissed):

hallvictoria (approved):

PR #410, opened 2026-02-05
Fix HttpResponseBodyInit to support Node.js Readable streams

### LINKED ISSUE #409: @azure/functions@4.11.1 broken HttpResponseBodyInit typings
Just updated from `@azure/functions@4.11.0` to `@azure/functions@4.11.1` and I get broken builds everywhere in all the samples using HTTP streaming:

```
src/functions/chats-post.ts:183:7 - error TS2322: Type 'Readable' is not assignable to type 'HttpResponseBodyInit'.
  Type 'Readable' is missing the following properties from type 'FormData': append, set, get, getAll, and 7 more.

183       body: jsonStream,
          ~~~~

  ../../node_modules/@azure/functions/types/http.d.ts:229:5
    229     body?: HttpResponseBodyInit;
            ~~~~
    The expected type comes from property 'body' which is declared here on type 'HttpResponseInit'
```

The `HttpResponseBodyInit` was previously defined as `any`, while on the new version it's:

```ts
export type HttpResponseBodyInit =
    | ReadableStream
    | Blob
    | ArrayBufferView
    | ArrayBuffer
    | FormData
    | URLSearchParams
    | string
    | null
    | undefined;
```

The type aliasing isn't working correctly in this case (not sure why) but this change is breaking many of our samples!

TsuyoshiUshio:
## Summary

Fixes #409 - `@azure/functions@4.11.1` breaks HTTP streaming with Node.js `Readable` streams.

## Root Cause Analysis

PR #405 (commit `81f6e9a`) removed the `lib.DOM` dependency by replacing the `BodyInit` type with a locally defined `HttpResponseBodyInit` type. However, this new type definition only included Web API types (`ReadableStream`, `Blob`, `ArrayBuffer`, etc.) and **did not include Node.js native `Readable` streams**.

This caused TypeScript compilation errors for users who pass Node.js `Readable` streams as HTTP response bodies:

```
error TS2322: Type 'Readable' is not assignable to type 'HttpResponseBodyInit'.
```

**Why this worked in v4.11.0**: In v4.11.0, `body` was typed as `BodyInit` from lib.DOM. However, in user projects without lib.DOM enabled, `BodyInit` resolved to `any`, allowing any type including `Readable`. The v4.11.1 change to an explicit union type surfaced this incompatibility.

**Why this is type-only fix**: Testing confirms that Node.js 20+ `Response` constructor natively accepts Node.js `Readable` streams (via undici), so no runtime conversion is needed.

## Changes

**types/http.d.ts**:
- Added `Readable` from `stream` module to imports
- Added `Readable` to the `HttpResponseBodyInit` union type
- Updated JSDoc comment

## Testing

- All 230 existing unit tests pass
- Verified Node.js `Response` accepts `Readable` directly: `new Response(readable)` works in Node.js 20+

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@azure/functions",
-    "version": "4.11.1",
+    "version": "4.11.2",
     "description": "Microsoft Azure Functions NodeJS Framework",
     "keywords": [
         "azure",
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,6 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License.
 
-export const version = '4.11.1';
+export const version = '4.11.2';
 
 export const returnBindingKey = '$return';
--- a/test/http/HttpResponse.test.ts
+++ b/test/http/HttpResponse.test.ts
@@ -6,6 +6,7 @@ import { Blob } from 'buffer';
 import * as chai from 'chai';
 import { expect } from 'chai';
 import * as chaiAsPromised from 'chai-as-promised';
+import { Readable } from 'stream';
 import { ReadableStream } from 'stream/web';
 import { HttpResponse } from '../../src/http/HttpResponse';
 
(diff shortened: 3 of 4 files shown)


swapnil-nagar:
Add the tests to fix this and update the package.json version

TsuyoshiUshio:
Thanks for the review! I've addressed the feedback:

1. ✅ Added tests for Node.js Readable stream body support (2 test cases)
2. ✅ Updated package.json version to 4.11.2
3. ✅ Fixed lint error (changed async generator to sync generator since no await was needed)

swapnil-nagar (approved):

PR #413, opened 2026-02-06
Adding the AsyncIterable type

swapnil-nagar:
PR: Add AsyncIterable<Uint8Array> support to HttpResponseBodyInit

Description:
Adds AsyncIterable<Uint8Array> as a valid body type for HTTP responses. This enables streaming response scenarios such as AI chat completions, where data is returned incrementally via async generators. Customers can now pass async iterables directly to HttpResponseInit.body without needing to convert them to ReadableStream first.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@azure/functions",
-    "version": "4.11.1",
+    "version": "4.11.2",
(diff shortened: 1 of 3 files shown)


TsuyoshiUshio (approved):
LGTM. Thank you for finding gap for undici.

PR #425, opened 2026-04-14
Adding the Structure content changes

swapnil-nagar:
https://github.com/Azure/azure-functions-nodejs-worker/issues/785

Adds structured content support for MCP Tool triggers.

Functions can now return typed content blocks (TextContentBlock, ImageContentBlock, AudioContentBlock, ResourceLinkContentBlock, CallToolResult) or use the new @McpContent class decorator to automatically emit both text and structuredContent from plain class instances.

Binary payloads in image/audio blocks are auto-encoded to base64. 

The host contract now always advertises useResultSchema: true, with optional resultSchema JSON Schema passthrough. Includes full unit test coverage across the conversion pipeline, marker utility, trigger contract, and InvocationModel serialization path (56 tests passing).

### DIFF
--- a/src/converters/toMcpToolResult.ts
+++ b/src/converters/toMcpToolResult.ts
@@ -1,0 +1,261 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import type { McpToolResult } from '@azure/functions';
+import { shouldCreateStructuredContentMarker } from '../utils/mcpContentMarker';
+
+interface TextContentBlock {
+    type: 'text';
+    text: string;
+}
+
+const multiContentResultType = 'multi_content_result';
+const textContentResultType = 'text';
+const directContentBlockTypes = new Set(['text', 'image', 'audio', 'resource', 'resource_link']);
+
+interface CallToolResultLike {
+    content: unknown[];
+    structuredContent?: unknown;
+}
+
+/**
+ * Converts a tool return value into a normalized MCP tool result.
+ *
+ * Supported inputs include CallToolResult-like objects, direct content blocks,
+ * arrays of content blocks, existing MCP results, and plain values.
+ */
+export function toMcpToolResult(result: unknown): McpToolResult | null | undefined {
+    if (result === null || result === undefined) {
+        return result;
+    }
+
+    if (isCallToolResult(result)) {
+        const callToolResult = ensureCallToolResultHasTextContent(result);
+        const normalizedBlocks = normalizeContentBlocks(callToolResult.content);
+        const mcpResult = createMcpToolResultFromContentBlocks(normalizedBlocks);
+        mcpResult.structuredContent = serializeStructuredContent(callToolResult.structuredContent);
+        return mcpResult;
+    }
+
+    if (isContentBlockArray(result)) {
+        const normalizedBlocks = normalizeContentBlocks(result);
+        const mcpResult = createMcpToolResultFromContentBlocks(normalizedBlocks);
+        return mcpResult;
+    }
+
+    if (isContentBlock(result)) {
+        const normalizedBlock = normalizeContentBlock(result);
+        const mcpResult = createMcpToolResultFromContentBlocks([normalizedBlock]);
+        return mcpResult;
+    }
+
+    if (isMcpToolResult(result)) {
+        return normalizeMcpToolResult(result);
+    }
+
+    const text = typeof result === 'string' ? result : JSON.stringify(result);
+    const mcpResult: McpToolResult = {
+        type: textContentResultType,
+        content: JSON.stringify({ type: textContentResultType, text } as TextContentBlock),
+    };
+
+    if (shouldCreateStructuredContentMarker(result)) {
+        mcpResult.structuredContent = JSON.stringify(result);
+    }
+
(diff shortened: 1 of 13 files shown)


hallvictoria (approved):

types/mcpTool.d.ts
```diff
@@ -132,3 +150,86 @@ export type Arg = Omit<McpToolProperty, 'propertyName'>;
  * Tool properties format - an object mapping property names to their definitions
  */
 export type Args = Record<string, Arg>;
+
+/**
+ * A text content block in an MCP tool response.
+ * Equivalent to .NET's TextContentBlock.
+ */
+export interface TextContentBlock {
```
liliankasem, line 158: This is concerning to me, defining your own interfaces for types that are exported by the MCP SDK is not a good idea. It will confuse customers, and it is already diverged from the MCP spec. There are also many risks to this:

   1. As the MCP spec evolves, these hand-rolled types won't track upstream changes automatically.
   2. Users already using @modelcontextprotocol/sdk can't pass SDK-typed objects directly 
  without casting.
   3. Semantic mismatches; Subtle differences (e.g., ResourceLink structure) could cause runtime issues when the host communicates with spec-compliant MCP clients.

I would expect you to take a deps on `@modelcontextprotocol/sdk`; the only reason we did not do this originally in the dotnet worker was because it was still in preview, so we went added a second pkg that brought in the MCP SDK: https://github.com/Azure/azure-functions-mcp-extension/blob/a76da5854342c95270c6b0cda50d33cb8f585736/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/Worker.Extensions.Mcp.Sdk.csproj#L10

But now that the MCP SDK is GA, it should be safe to depend on it.

src/converters/toMcpToolResult.ts
```diff
@@ -0,0 +1,261 @@
...
+    const text = typeof result === 'string' ? result : JSON.stringify(result);
+    const mcpResult: McpToolResult = {
+        type: textContentResultType,
+        content: JSON.stringify({ type: textContentResultType, text } as TextContentBlock),
+    };
+
+    if (shouldCreateStructuredContentMarker(result)) {
+        mcpResult.structuredContent = JSON.stringify(result);
+    }
+
```
liliankasem, line 65: I think this approach is really fragile. The detection chain you have here to identify the mcp tool result type can break in several ways:

1. isCallToolResult is far too loose — it only checks Array.isArray(value.content):

 function isCallToolResult(value: unknown): value is CallToolResultLike {
     return !!value && typeof value === 'object' && Array.isArray((value as any).content);
 }

Any object with a content array matches. A user returning { content: ["hello"], status: "ok" } would be
misinterpreted as a CallToolResult and processed through the content-block pipeline.

2. isContentBlock vs isMcpToolResult overlap — Both check for .type as a string. The only differentiator is
that isContentBlock excludes objects with a .content property. But isMcpToolResult accepts objects with 
.structuredContent and no .content. So an object like:

 { type: 'text', structuredContent: '...' }

Would fail isContentBlock (type is in the set, no .content — actually this would pass isContentBlock since it
doesn't have .content). The ordering saves you here, but only by accident.

3. The directContentBlockTypes set is a hardcoded allowlist:

 const directContentBlockTypes = new Set(['text', 'image', 'audio', 'resource', 'resource_link']);

When the MCP spec adds new content block types, this silently falls through to the plain-value path,
double-JSON-encoding the result.

4. isMcpToolResult catches too much — anything with { type: string, content: ... } or { type: string,
structuredContent: ... } matches. A user's domain object like { type: 'report', content: 'quarterly results' }
would be treated as a pre-built MCP result.

With my suggestion for using the MCP SDK, you can instead import the SDK types and use `instanceof` checks against them.

types/mcpTool.d.ts
swapnil-nagar: I understand the concern about defining our own interfaces for types that already exist in the MCP SDK. I agree this can lead to risks like drifting from the spec, subtle mismatches, and extra maintenance as the MCP spec changes over time.

That said, adding a direct dependency on @modelcontextprotocol/sdk comes with its own trade-offs. It’s a fairly heavy package (for example, it pulls in dependencies like zod), and adding it to the core library would increase the bundle size for all users even those who don’t use MCP. This would directly impact cold start performance, which is an important metric for us.

Recently, we’ve worked on reducing the size of @azure/functions, which led to noticeable improvements (around ~2.2s at p99 and ~60ms at p50). Bringing in a large dependency would undo some of that progress and affect all customers, not just those using MCP.

While I agree that redefining types has long-term risks, adding a heavy dependency at the core level has a broader impact on the platform. 

**_It’s better to keep the core package lightweight based on the tradeoff._**

src/converters/toMcpToolResult.ts
swapnil-nagar: How about I provided a classes from @azure/functions and tell users to return one of them when they want MCP semantics.
 
Benefit would we would are enforcing our own programming model which is more controllable and customer needs to adhere to our return types. Another benefit is zero runtime deps(@MCP/SDK), and zero ambiguity. No confusion with CallToolResult. 

https://github.com/Azure/azure-functions-nodejs-library/pull/427

PR #427, opened 2026-04-16
Replace structural MCP tool-result detection with class-based API

swapnil-nagar:
Unambiguous detection via instanceof,  no more structural guessing. [{ content: [...], status: "ok" }](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) or [{ type: 'report', content: '...' }](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) are definitively plain user values, eliminating all four false-positive paths you identified.

Zero runtime cost vs. the SDK — no @modelcontextprotocol/sdk or Zod dependency, so cold start on Azure Functions workers stays unaffected. The classes are small and tree-shakeable.

Self-normalizing payloads — each block's [toJSON()](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) handles Buffer/ArrayBuffer → base64 and omits undefined fields, so the converter no longer has to walk unknown object shapes looking for [data](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fields to encode.

Discoverable, typed API — IntelliSense surfaces [TextContent](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [ImageContent](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [McpToolResponse](vscode-file://vscode-app/c:/Users/swapnilnagar/AppData/Local/Programs/Microsoft%20VS%20Code/ae130017f8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), etc. with typed constructors. Users can't accidentally produce a malformed block the way they could with raw object literals.

Simpler, smaller converter — the detection chain collapsed from five overlapping predicates plus a fallback to three instanceof checks, making the code easier to audit and less likely to regress.

Example would be 

```

export class McpToolResponse {
    readonly content: McpContentBlock[];
    readonly structuredContent?: unknown;
    readonly isError?: boolean;
}

```

usage would be instead of 

```
export async function getImageWithMetadata(
    _toolArguments: unknown,
    _context: InvocationContext
): Promise<CallToolResult> {
    const metadata = {
        imageId: 'logo',
        format: 'png',
        tags: ['functions', 'azure', 'serverless'],
    };

    const content: [TextContentBlock, ImageContentBlock] = [
        { type: 'text', text: JSON.stringify(metadata) },
        { type: 'image', data: base64ImageData, mimeType: 'image/png' },
    ];

    return {
        content,
        structuredContent: metadata,
    };
}
```

The contract would be cleaner

```
export async function getImageWithMetadata(
    _toolArguments: unknown,
    _context: InvocationContext
): Promise<McpToolResponse> {
    const metadata = {
        imageId: 'logo',
        format: 'png',
        tags: ['functions', 'azure', 'serverless'],
    };

    return new McpToolResponse({
        content: [
            new TextContent(JSON.stringify(metadata)),
            new ImageContent({ data: base64ImageData, mimeType: 'image/png' }),
        ],
        structuredContent: metadata,
    });
}
```

### DIFF
--- a/src/mcp/McpToolResponse.ts
+++ b/src/mcp/McpToolResponse.ts
@@ -29,6 +29,90 @@ function normalizeBinaryData(data: BinaryData): string {
  * Base class for all MCP content blocks. The library uses `instanceof McpContentBlock`
  * to discriminate content blocks from plain user values, so every content-block subclass
  * must extend this class.
+ *
+ * ## Extending with custom content block types
+ *
+ * The library ships built-in subclasses for the MCP spec's current content block types
+ * (`TextContent`, `ImageContent`, `AudioContent`, `ResourceLinkContent`, `ResourceContent`).
+ * If the spec adds a new block type — or your scenario needs a custom one — you can ship
+ * your own subclass without any library change. The converter only checks
+ * `instanceof McpContentBlock` and calls `JSON.stringify(block)`, which invokes your
+ * `toJSON()` to produce the wire payload.
+ *
+ * ### Example: adding a hypothetical `VideoContent`
+ *
+ * ```ts
+ * import { McpContentBlock } from '@azure/functions';
+ *
+ * export interface VideoContentInit {
+ *     data: string | Buffer | ArrayBuffer;
+ *     mimeType: string;          // e.g. 'video/mp4'
+ *     durationMs?: number;       // optional field from a hypothetical spec
+ * }
+ *
+ * export class VideoContent extends McpContentBlock {
+ *     readonly type = 'video' as const;
+ *     readonly data: string | Buffer | ArrayBuffer;
+ *     readonly mimeType: string;
+ *     readonly durationMs?: number;
+ *
+ *     constructor(init: VideoContentInit) {
+ *         super();
+ *         this.data = init.data;
+ *         this.mimeType = init.mimeType;
+ *         this.durationMs = init.durationMs;
+ *     }
+ *
+ *     toJSON(): Record<string, unknown> {
+ *         const out: Record<string, unknown> = {
+ *             type: this.type,
+ *             data: toBase64(this.data),
+ *             mimeType: this.mimeType,
+ *         };
+ *         if (this.durationMs !== undefined) out.durationMs = this.durationMs;
+ *         return out;
+ *     }
+ * }
+ *
+ * function toBase64(data: string | Buffer | ArrayBuffer): string {
+ *     if (typeof data === 'string') return data;
+ *     if (Buffer.isBuffer(data)) return data.toString('base64');
+ *     return Buffer.from(new Uint8Array(data)).toString('base64');
+ * }
+ * ```
+ *
+ * ### Using a custom block from a tool handler
+ *
+ * Return it standalone, or mix it with built-in blocks inside an `McpToolResponse`:
+ *
+ * ```ts
+ * // Single block
+ * handler: async () => new VideoContent({ data: buf, mimeType: 'video/mp4' })
+ *
+ * // Mixed with built-ins + structured content
+ * handler: async () => new McpToolResponse({
+ *     content: [
+ *         new TextContent('Detected 3 scenes'),
+ *         new VideoContent({ data: buf, mimeType: 'video/mp4' }),
+ *     ],
+ *     structuredContent: { scenes: 3, confidence: 0.92 },
+ * })
+ * ```
+ *
+ * ### What the library does for you
+ *
+ *  - `instanceof McpContentBlock` treats your subclass identically to built-in blocks.
+ *  - Single-block returns propagate your `type` string to the outer result; arrays are
+ *    wrapped as `multi_content_result`.
+ *  - `structuredContent` handling, fallback-text synthesis, and nullish passthrough all
+ *    apply unchanged.
+ *
+ * ### What you must get right in your subclass
+ *
+ *  - `toJSON()` must return the exact wire shape the spec requires for your `type`.
+ *  - Binary payloads should be base64-encoded in `toJSON()` (see the `toBase64` helper
+ *    above).
+ *  - Plain object literals are **not** recognized — you must construct an instance.
  */
 export abstract class McpContentBlock {
     abstract readonly type: string;
--- a/src/mcp/McpToolResponse.ts
+++ b/src/mcp/McpToolResponse.ts
@@ -2,11 +2,11 @@
 // Licensed under the MIT License.
 
 import type {
-    AudioContentInit as IAudioContentInit,
-    ImageContentInit as IImageContentInit,
+    McpAudioContentInit as IAudioContentInit,
+    McpImageContentInit as IImageContentInit,
+    McpResourceContentInit as IResourceContentInit,
+    McpResourceLinkContentInit as IResourceLinkContentInit,
     McpToolResponseInit as IMcpToolResponseInit,
-    ResourceContentInit as IResourceContentInit,
-    ResourceLinkContentInit as IResourceLinkContentInit,
 } from '@azure/functions';
 
 type BinaryData = string | Buffer | ArrayBuffer;
@@ -33,7 +33,7 @@ function normalizeBinaryData(data: BinaryData): string {
  * ## Extending with custom content block types
  *
  * The library ships built-in subclasses for the MCP spec's current content block types
- * (`TextContent`, `ImageContent`, `AudioContent`, `ResourceLinkContent`, `ResourceContent`).
+ * (`McpTextContent`, `McpImageContent`, `McpAudioContent`, `McpResourceLinkContent`, `McpResourceContent`).
  * If the spec adds a new block type — or your scenario needs a custom one — you can ship
  * your own subclass without any library change. The converter only checks
  * `instanceof McpContentBlock` and calls `JSON.stringify(block)`, which invokes your
@@ -92,7 +92,7 @@ function normalizeBinaryData(data: BinaryData): string {
  * // Mixed with built-ins + structured content
  * handler: async () => new McpToolResponse({
  *     content: [
- *         new TextContent('Detected 3 scenes'),
+ *         new McpTextContent('Detected 3 scenes'),
  *         new VideoContent({ data: buf, mimeType: 'video/mp4' }),
  *     ],
  *     structuredContent: { scenes: 3, confidence: 0.92 },
@@ -124,7 +124,7 @@ export abstract class McpContentBlock {
     abstract toJSON(): Record<string, unknown>;
 }
 
-export class TextContent extends McpContentBlock {
+export class McpTextContent extends McpContentBlock {
     readonly type = 'text' as const;
     readonly text: string;
 
@@ -138,7 +138,7 @@ export class TextContent extends McpContentBlock {
     }
 }
 
-export class ImageContent extends McpContentBlock {
+export class McpImageContent extends McpContentBlock {
     readonly type = 'image' as const;
     readonly data: BinaryData;
     readonly mimeType?: string;
@@ -161,7 +161,7 @@ export class ImageContent extends McpContentBlock {
     }
 }
 
-export class AudioContent extends McpContentBlock {
+export class McpAudioContent extends McpContentBlock {
     readonly type = 'audio' as const;
     readonly data: BinaryData;
     readonly mimeType?: string;
@@ -184,7 +184,7 @@ export class AudioContent extends McpContentBlock {
     }
 }
 
-export class ResourceLinkContent extends McpContentBlock {
+export class McpResourceLinkContent extends McpContentBlock {
     readonly type = 'resource_link' as const;
     readonly uri: string;
     readonly name?: string;
@@ -208,7 +208,7 @@ export class ResourceLinkContent extends McpContentBlock {
     }
 }
 
-export class ResourceContent extends McpContentBlock {
+export class McpResourceContent extends McpContentBlock {
     readonly type = 'resource' as const;
     readonly resource: IResourceContentInit['resource'];
 
--- a/src/mcp/McpToolResponse.ts
+++ b/src/mcp/McpToolResponse.ts
@@ -1,0 +1,161 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
+import type {
+    AudioContentInit as IAudioContentInit,
+    ImageContentInit as IImageContentInit,
+    McpToolResponseInit as IMcpToolResponseInit,
+    ResourceContentInit as IResourceContentInit,
+    ResourceLinkContentInit as IResourceLinkContentInit,
+} from '@azure/functions';
+
+type BinaryData = string | Buffer | ArrayBuffer;
+
+function normalizeBinaryData(data: BinaryData): string {
+    if (typeof data === 'string') {
+        return data;
+    }
+    if (Buffer.isBuffer(data)) {
+        return data.toString('base64');
+    }
+    if (ArrayBuffer.isView(data)) {
+        const view = new Uint8Array(data.buffer, data.byteOffset, data.byteLength);
+        return Buffer.from(view).toString('base64');
+    }
+    return Buffer.from(new Uint8Array(data)).toString('base64');
+}
+
+/**
+ * Base class for all MCP content blocks. The library uses `instanceof McpContentBlock`
+ * to discriminate content blocks from plain user values, so every content-block subclass
+ * must extend this class.
+ */
+export abstract class McpContentBlock {
+    abstract readonly type: string;
+
+    /**
+     * Returns the wire representation of this block. Subclasses override to normalize
+     * binary payloads and omit undefined fields.
+     */
+    abstract toJSON(): Record<string, unknown>;
+}
+
+export class TextContent extends McpContentBlock {
+    readonly type = 'text' as const;
+    readonly text: string;
+
+    constructor(text: string) {
+        super();
+        this.text = text;
+    }
+
+    toJSON(): Record<string, unknown> {
+        return { type: this.type, text: this.text };
+    }
+}
+
+export class ImageContent extends McpContentBlock {
+    readonly type = 'image' as const;
+    readonly data: BinaryData;
+    readonly mimeType?: string;
+
+    constructor(init: IImageContentInit) {
+        super();
+        this.data = init.data;
+        this.mimeType = init.mimeType;
+    }
+
+    toJSON(): Record<string, unknown> {
+        const out: Record<string, unknown> = {
+            type: this.type,
+            data: normalizeBinaryData(this.data),
+        };
+        if (this.mimeType !== undefined) {
+            out.mimeType = this.mimeType;
+        }
+        return out;
+    }
+}
+
+export class AudioContent extends McpContentBlock {
+    readonly type = 'audio' as const;
+    readonly data: BinaryData;
+    readonly mimeType?: string;
+
+    constructor(init: IAudioContentInit) {
+        super();
+        this.data = init.data;
+        this.mimeType = init.mimeType;
+    }
+
+    toJSON(): Record<string, unknown> {
+        const out: Record<string, unknown> = {
+            type: this.type,
+            data: normalizeBinaryData(this.data),
+        };
+        if (this.mimeType !== undefined) {
+            out.mimeType = this.mimeType;
+        }
+        return out;
+    }
+}
+
+export class ResourceLinkContent extends McpContentBlock {
+    readonly type = 'resource_link' as const;
+    readonly uri: string;
+    readonly name?: string;
+    readonly description?: string;
+    readonly mimeType?: string;
+
+    constructor(init: IResourceLinkContentInit) {
+        super();
+        this.uri = init.uri;
+        this.name = init.name;
+        this.description = init.description;
+        this.mimeType = init.mimeType;
+    }
+
+    toJSON(): Record<string, unknown> {
+        const out: Record<string, unknown> = { type: this.type, uri: this.uri };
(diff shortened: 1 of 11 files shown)


hallvictoria (dismissed):

liliankasem:
This is a good improvement over $425, esp given the constraints. But there is still one more gap I think we should consider. Do we want to support customers being able to use `@modelcontextprotocol/sdk` types in their Function App? E.g.

```typescript
import type { CallToolResult } from '@modelcontextprotocol/sdk';

app.mcpTool('myTool', {
    handler: async (): Promise<CallToolResult> => ({
        content: [{ type: 'text', text: 'hello' }],
        structuredContent: { id: 'x1' }
    })
});
```

With this PR they have to use the types you have defined (which also means you will now have to maintain these can keep them up to date with the spec), and this return value fails all three `instanceof` checks and falls through to the plain-text path (which makes sense since your code has no awareness of the types from `@modelcontextprotocol/sdk`. 

You could introduce a fallback after `instanceof` steps to see if something "looks like" the models from the mcp sdk, before we fallback to the plain text? Or, we may wantto warn users that MCP sdk types are not supported.

src/mcp/McpToolResponse.ts
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: What's going to be the process for keeping this up to date with the spec?

TsuyoshiUshio:
Thanks for this — the class-based detection is a big readability/correctness win. I have a few suggestions before we merge.

## Main concern: Breaking change without a deprecation window

This PR removes publicly exported types that shipped in v4.11.x:
- `TextContentBlock`, `ImageContentBlock`, `AudioContentBlock`, `ResourceLinkContentBlock`
- `CallToolResult`
- `McpContentBlock` as a **union type** (now an abstract class with the same name — breaks code that used it as `McpContentBlock[]` union of literals)

Users who imported these types, or who returned plain object literals like
`{ type: 'text', text: '...' }` or `{ content: [...], structuredContent: {...} }`,
will either get a compile error or a **silent behavior change**: their object will now
be serialized as a JSON-stringified text block instead of being treated as MCP content.

Since v4 is GA and these types are part of the public surface, could we do this in two
steps? Concretely:

### Step 1 — v4.next (minor, non-breaking)

**`src/converters/toMcpToolResult.ts`** — keep both detectors, new API wins:

```ts
export function toMcpToolResult(result: unknown): McpToolResult | null | undefined {
    if (result === null || result === undefined) {
        return result;
    }

    // --- New class-based API (preferred, unambiguous) ---
    if (result instanceof McpToolResponse) {
        return serializeToolResponse(result);
    }
    if (result instanceof McpContentBlock) {
        return serializeToolResponse(new McpToolResponse({ content: [result] }));
    }
    if (Array.isArray(result) && result.length > 0 && result.every((b) => b instanceof McpContentBlock)) {
        return serializeToolResponse(new McpToolResponse({ content: result as McpContentBlock[] }));
    }

    // --- Legacy structural detection (kept for back-compat, @deprecated) ---
    // TODO(v5): remove this block and the helpers below.
    if (isCallToolResult(result)) {
        const callToolResult = ensureCallToolResultHasTextContent(result);
        const normalizedBlocks = normalizeContentBlocks(callToolResult.content);
        const mcpResult = createMcpToolResultFromContentBlocks(normalizedBlocks);
        mcpResult.structuredContent = serializeStructuredContent(callToolResult.structuredContent);
        return mcpResult;
    }
    if (isContentBlockArray(result)) {
        const normalizedBlocks = normalizeContentBlocks(result);
        return createMcpToolResultFromContentBlocks(normalizedBlocks);
    }
    if (isContentBlock(result)) {
        const normalizedBlock = normalizeContentBlock(result);
        return createMcpToolResultFromContentBlocks([normalizedBlock]);
    }
    if (isMcpToolResult(result)) {
        return normalizeMcpToolResult(result);
    }

    // --- Fallback: serialize as text ---
    const text = typeof result === 'string' ? result : JSON.stringify(result);
    const mcpResult: McpToolResult = {
        type: textContentResultType,
        content: JSON.stringify({ type: textContentResultType, text }),
    };
    if (shouldCreateStructuredContentMarker(result)) {
        mcpResult.structuredContent = JSON.stringify(result);
    }
    return mcpResult;
}
```

**`types/mcpTool.d.ts`** — keep the old interfaces as `@deprecated`, add the new classes alongside:

```ts
// --- New class-based API ---
export declare abstract class McpContentBlock {
    abstract readonly type: string;
    abstract toJSON(): Record<string, unknown>;
}

export declare class TextContent extends McpContentBlock {
    readonly type: 'text';
    readonly text: string;
    constructor(text: string);
    toJSON(): Record<string, unknown>;
}

// ... ImageContent, AudioContent, ResourceLinkContent, ResourceContent ...

export declare class McpToolResponse {
    readonly content: McpContentBlock[];
    readonly structuredContent?: unknown;
    readonly isError?: boolean;
    constructor(init: McpToolResponseInit);
}

// --- Deprecated legacy types (removed in v5) ---

/**
 * @deprecated Use the `TextContent` class instead. Object literals are detected
 * via structural matching and can be ambiguous with user domain objects.
 */
export interface TextContentBlock {
    type: 'text';
    text: string;
}

/** @deprecated Use the `ImageContent` class instead. */
export interface ImageContentBlock {
    type: 'image';
    data: string | Buffer | ArrayBuffer;
    mimeType?: string;
}

// ... AudioContentBlock, ResourceLinkContentBlock (same pattern) ...

/** @deprecated Use the `McpToolResponse` class instead. */
export interface CallToolResult {
    content: McpContentBlockUnion[];
    structuredContent?: unknown;
    isError?: boolean;
}

/** @deprecated Union retained for back-compat. Prefer the `McpContentBlock` class hierarchy. */
export type McpContentBlockUnion =
    | TextContentBlock
    | ImageContentBlock
    | AudioContentBlock
    | ResourceLinkContentBlock
    | Record<string, unknown>;
```

> Naming note: the old `McpContentBlock` **union** clashes with the new `McpContentBlock`
> **class**. One option is to rename the deprecated union to `McpContentBlockUnion` (as
> above). If that itself is considered a break, an alternative is to name the new class
> something like `McpContentBlockBase` and keep the existing union as-is for v4.

**CHANGELOG / README:** add a migration note pointing users to the new classes.

### Step 2 — v5.0 (major, later)

Remove the legacy detectors, the deprecated interfaces, and the rename shim. That's the
cleaned-up state this PR targets today — just done in a follow-up once users have had a
release or two to migrate.

This gives existing users a clear migration path and lets the new API win on its own
merits without tying the correctness improvement to a breaking release.

---

## Other items (independent of the break/non-break decision)

### 1. `isError` is advertised but dropped on the wire

`McpToolResponse` exposes `isError?: boolean`, but `serializeToolResponse` never copies
it onto the returned `McpToolResult`:

```ts
function serializeToolResponse(response: McpToolResponse): McpToolResult {
    const blocks = ensureTextBlockWhenStructured(response);
    // ...
    const out: McpToolResult = { type, content: contentStr };

    if (response.structuredContent !== undefined && response.structuredContent !== null) {
        out.structuredContent = /* ... */;
    }
    // response.isError is never propagated
    return out;
}
```

Fix:

```ts
if (response.isError !== undefined) {
    out.isError = response.isError;
}
```

This likely also requires adding `isError?: boolean` to the `McpToolResult` wire type
(if it isn't there already). Otherwise, we're advertising a feature in the public class
that silently no-ops, which is worse than not exposing it.

### 2. `ResourceContent` has no test

`TextContent`, `ImageContent`, `AudioContent`, and `ResourceLinkContent` each have a
dedicated serialization test in `toMcpToolResult.test.ts`, but `ResourceContent` doesn't.
Given it has the most complex `toJSON` (nested `resource` with base64 `blob`
normalization), it's worth covering. Suggested test:

```ts
it('serializes a ResourceContent block with base64-encoded blob', () => {
    const blob = Buffer.from('binary-data');
    const result = toMcpToolResult(
        new ResourceContent({
            resource: { uri: 'file:///a.bin', mimeType: 'application/octet-stream', blob },
        })
    );
    expect(result?.type).to.equal('resource');
    const content = JSON.parse(result?.content || '{}');
    expect(content).to.deep.equal({
        type: 'resource',
        resource: {
            uri: 'file:///a.bin',
            mimeType: 'application/octet-stream',
            blob: blob.toString('base64'),
        },
    });
});

it('serializes a ResourceContent block with only text', () => {
    const result = toMcpToolResult(
        new ResourceContent({ resource: { uri: 'file:///a.txt', text: 'hello' } })
    );
    const content = JSON.parse(result?.content || '{}');
    expect(content.resource).to.deep.equal({ uri: 'file:///a.txt', text: 'hello' });
    expect(content.resource).to.not.have.property('blob');
    expect(content.resource).to.not.have.property('mimeType');
});
```

### 3. Minor: lock in the empty-array case

`toMcpToolResult([])` falls through to the plain-value branch and becomes
`JSON.stringify([])` as a text block. That matches prior behavior, but a one-liner
regression test is cheap:

```ts
it('treats an empty array as a plain value', () => {
    const result = toMcpToolResult([]);
    expect(result?.type).to.equal('text');
    const content = JSON.parse(result?.content || '{}');
    expect(content.text).to.equal('[]');
});
```

---

Happy to help with any of the above if useful. Otherwise the direction looks great.

src/mcp/McpToolResponse.ts
```diff
@@ -0,0 +1,161 @@
...
+        super();
+        this.text = text;
+    }
+
+    toJSON(): Record<string, unknown> {
+        return { type: this.type, text: this.text };
+    }
+}
+
+export class ImageContent extends McpContentBlock {
```
liliankasem: I think all Mcp types going into @azure/functions needs to be prefixed with Mcp


```suggestion
export class McpImageContent extends McpContentBlock {
```

src/mcp/McpToolResponse.ts
swapnil-nagar: We need to maintain the spec as a tradeoff in absence of @mcp/types library (planning to raise a issue with @mcp repo).

We will create a mechanism to support different specs by regular scanning the specs. Also adding the warning in next commit, so we can periodically scan it to observe the logs and plan ahead.

We own the MCP wire format and maintain the azure functions programming model. Hence we will not introduce the adapter/fallback.
But, If someone has SDK shapes, they can map them to our classes in a one-off helper in their app. We don't need to own that conversion.
The warning alone may be enough signal. Many users will just construct our classes directly after seeing the warning once.

src/mcp/McpToolResponse.ts
swapnil-nagar: Fixed

swapnil-nagar:
> Thanks for this — the class-based detection is a big readability/correctness win. I have a few suggestions before we merge.
> 
> ## Main concern: Breaking change without a deprecation window
> This PR removes publicly exported types that shipped in v4.11.x:
> 
> * `TextContentBlock`, `ImageContentBlock`, `AudioContentBlock`, `ResourceLinkContentBlock`
> * `CallToolResult`
> * `McpContentBlock` as a **union type** (now an abstract class with the same name — breaks code that used it as `McpContentBlock[]` union of literals)
> 
> Users who imported these types, or who returned plain object literals like `{ type: 'text', text: '...' }` or `{ content: [...], structuredContent: {...} }`, will either get a compile error or a **silent behavior change**: their object will now be serialized as a JSON-stringified text block instead of being treated as MCP content.
> 
> Since v4 is GA and these types are part of the public surface, could we do this in two steps? Concretely:
> 
> ### Step 1 — v4.next (minor, non-breaking)
> **`src/converters/toMcpToolResult.ts`** — keep both detectors, new API wins:
> 
> ```ts
> export function toMcpToolResult(result: unknown): McpToolResult | null | undefined {
>     if (result === null || result === undefined) {
>         return result;
>     }
> 
>     // --- New class-based API (preferred, unambiguous) ---
>     if (result instanceof McpToolResponse) {
>         return serializeToolResponse(result);
>     }
>     if (result instanceof McpContentBlock) {
>         return serializeToolResponse(new McpToolResponse({ content: [result] }));
>     }
>     if (Array.isArray(result) && result.length > 0 && result.every((b) => b instanceof McpContentBlock)) {
>         return serializeToolResponse(new McpToolResponse({ content: result as McpContentBlock[] }));
>     }
> 
>     // --- Legacy structural detection (kept for back-compat, @deprecated) ---
>     // TODO(v5): remove this block and the helpers below.
>     if (isCallToolResult(result)) {
>         const callToolResult = ensureCallToolResultHasTextContent(result);
>         const normalizedBlocks = normalizeContentBlocks(callToolResult.content);
>         const mcpResult = createMcpToolResultFromContentBlocks(normalizedBlocks);
>         mcpResult.structuredContent = serializeStructuredContent(callToolResult.structuredContent);
>         return mcpResult;
>     }
>     if (isContentBlockArray(result)) {
>         const normalizedBlocks = normalizeContentBlocks(result);
>         return createMcpToolResultFromContentBlocks(normalizedBlocks);
>     }
>     if (isContentBlock(result)) {
>         const normalizedBlock = normalizeContentBlock(result);
>         return createMcpToolResultFromContentBlocks([normalizedBlock]);
>     }
>     if (isMcpToolResult(result)) {
>         return normalizeMcpToolResult(result);
>     }
> 
>     // --- Fallback: serialize as text ---
>     const text = typeof result === 'string' ? result : JSON.stringify(result);
>     const mcpResult: McpToolResult = {
>         type: textContentResultType,
>         content: JSON.stringify({ type: textContentResultType, text }),
>     };
>     if (shouldCreateStructuredContentMarker(result)) {
>         mcpResult.structuredContent = JSON.stringify(result);
>     }
>     return mcpResult;
> }
> ```
> 
> **`types/mcpTool.d.ts`** — keep the old interfaces as `@deprecated`, add the new classes alongside:
> 
> ```ts
> // --- New class-based API ---
> export declare abstract class McpContentBlock {
>     abstract readonly type: string;
>     abstract toJSON(): Record<string, unknown>;
> }
> 
> export declare class TextContent extends McpContentBlock {
>     readonly type: 'text';
>     readonly text: string;
>     constructor(text: string);
>     toJSON(): Record<string, unknown>;
> }
> 
> // ... ImageContent, AudioContent, ResourceLinkContent, ResourceContent ...
> 
> export declare class McpToolResponse {
>     readonly content: McpContentBlock[];
>     readonly structuredContent?: unknown;
>     readonly isError?: boolean;
>     constructor(init: McpToolResponseInit);
> }
> 
> // --- Deprecated legacy types (removed in v5) ---
> 
> /**
>  * @deprecated Use the `TextContent` class instead. Object literals are detected
>  * via structural matching and can be ambiguous with user domain objects.
>  */
> export interface TextContentBlock {
>     type: 'text';
>     text: string;
> }
> 
> /** @deprecated Use the `ImageContent` class instead. */
> export interface ImageContentBlock {
>     type: 'image';
>     data: string | Buffer | ArrayBuffer;
>     mimeType?: string;
> }
> 
> // ... AudioContentBlock, ResourceLinkContentBlock (same pattern) ...
> 
> /** @deprecated Use the `McpToolResponse` class instead. */
> export interface CallToolResult {
>     content: McpContentBlockUnion[];
>     structuredContent?: unknown;
>     isError?: boolean;
> }
> 
> /** @deprecated Union retained for back-compat. Prefer the `McpContentBlock` class hierarchy. */
> export type McpContentBlockUnion =
>     | TextContentBlock
>     | ImageContentBlock
>     | AudioContentBlock
>     | ResourceLinkContentBlock
>     | Record<string, unknown>;
> ```
> 
> > Naming note: the old `McpContentBlock` **union** clashes with the new `McpContentBlock`
> > **class**. One option is to rename the deprecated union to `McpContentBlockUnion` (as
> > above). If that itself is considered a break, an alternative is to name the new class
> > something like `McpContentBlockBase` and keep the existing union as-is for v4.
> 
> **CHANGELOG / README:** add a migration note pointing users to the new classes.
> 
> ### Step 2 — v5.0 (major, later)
> Remove the legacy detectors, the deprecated interfaces, and the rename shim. That's the cleaned-up state this PR targets today — just done in a follow-up once users have had a release or two to migrate.
> 
> This gives existing users a clear migration path and lets the new API win on its own merits without tying the correctness improvement to a breaking release.
> 
> ## Other items (independent of the break/non-break decision)
> ### 1. `isError` is advertised but dropped on the wire
> `McpToolResponse` exposes `isError?: boolean`, but `serializeToolResponse` never copies it onto the returned `McpToolResult`:
> 
> ```ts
> function serializeToolResponse(response: McpToolResponse): McpToolResult {
>     const blocks = ensureTextBlockWhenStructured(response);
>     // ...
>     const out: McpToolResult = { type, content: contentStr };
> 
>     if (response.structuredContent !== undefined && response.structuredContent !== null) {
>         out.structuredContent = /* ... */;
>     }
>     // response.isError is never propagated
>     return out;
> }
> ```
> 
> Fix:
> 
> ```ts
> if (response.isError !== undefined) {
>     out.isError = response.isError;
> }
> ```
> 
> This likely also requires adding `isError?: boolean` to the `McpToolResult` wire type (if it isn't there already). Otherwise, we're advertising a feature in the public class that silently no-ops, which is worse than not exposing it.
> 
> ### 2. `ResourceContent` has no test
> `TextContent`, `ImageContent`, `AudioContent`, and `ResourceLinkContent` each have a dedicated serialization test in `toMcpToolResult.test.ts`, but `ResourceContent` doesn't. Given it has the most complex `toJSON` (nested `resource` with base64 `blob` normalization), it's worth covering. Suggested test:
> 
> ```ts
> it('serializes a ResourceContent block with base64-encoded blob', () => {
>     const blob = Buffer.from('binary-data');
>     const result = toMcpToolResult(
>         new ResourceContent({
>             resource: { uri: 'file:///a.bin', mimeType: 'application/octet-stream', blob },
>         })
>     );
>     expect(result?.type).to.equal('resource');
>     const content = JSON.parse(result?.content || '{}');
>     expect(content).to.deep.equal({
>         type: 'resource',
>         resource: {
>             uri: 'file:///a.bin',
>             mimeType: 'application/octet-stream',
>             blob: blob.toString('base64'),
>         },
>     });
> });
> 
> it('serializes a ResourceContent block with only text', () => {
>     const result = toMcpToolResult(
>         new ResourceContent({ resource: { uri: 'file:///a.txt', text: 'hello' } })
>     );
>     const content = JSON.parse(result?.content || '{}');
>     expect(content.resource).to.deep.equal({ uri: 'file:///a.txt', text: 'hello' });
>     expect(content.resource).to.not.have.property('blob');
>     expect(content.resource).to.not.have.property('mimeType');
> });
> ```
> 
> ### 3. Minor: lock in the empty-array case
> `toMcpToolResult([])` falls through to the plain-value branch and becomes `JSON.stringify([])` as a text block. That matches prior behavior, but a one-liner regression test is cheap:
> 
> ```ts
> it('treats an empty array as a plain value', () => {
>     const result = toMcpToolResult([]);
>     expect(result?.type).to.equal('text');
>     const content = JSON.parse(result?.content || '{}');
>     expect(content.text).to.equal('[]');
> });
> ```
> 
> Happy to help with any of the above if useful. Otherwise the direction looks great.

@TsuyoshiUshio We have not release this change yet so no breaking change has introduced. 

On tests I have fixed it in last commit.

src/mcp/McpToolResponse.ts
liliankasem: I think it would be easy to setup a GH workflow or something to either open a PR or issue if there are any spec changes to these types

swapnil-nagar: image

We have added the warning and docs so that customer is informed.

Also, I will explore the right mechanism for specs.

src/mcp/sdkCompat.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Idempotent: subsequent calls are no-ops to avoid log spam.
+ */
+export function warnIfLooksLikeMcpSdkValue(value: unknown): void {
+    if (warned || !looksLikeMcpSdkValue(value)) {
+        return;
+    }
+    warned = true;
+    // eslint-disable-next-line no-console
+    console.warn(
```
liliankasem: Would suggest something a little more clear e.g.

```js
 console.warn(
     '[@azure/functions] Tool handler return value appears to use types from ' +
         '`@modelcontextprotocol/sdk` (e.g. `CallToolResult`, `TextContent`, `ImageContent`). ' +
         'These types are not supported directly and will be serialized as plain text. ' +
         'Use the equivalent classes from `@azure/functions` instead — ' +
         '`McpToolResponse`, `McpTextContent`, `McpImageContent`, etc. ' +
         'See the `McpContentBlock` JSDoc for custom content block examples.'
 );
```

liliankasem (dismissed):

src/mcp/sdkCompat.ts
swapnil-nagar: Sure, I have addressed this comment and using context logger for readable colored warning.

hallvictoria (approved):

PR #431, opened 2026-04-29
Fix Web PubSub bindings

Y-Sindo:
The `connection` property should be `connections`.

### DIFF
--- a/src/trigger.ts
+++ b/src/trigger.ts
@@ -129,8 +129,11 @@ export function mySql(options: MySqlTriggerOptions): MySqlTrigger {
 }
 
 export function webPubSub(options: WebPubSubTriggerOptions): WebPubSubTrigger {
+    // eslint-disable-next-line deprecation/deprecation
+    const { connection, ...rest } = options;
     return addTriggerBindingName({
-        ...options,
+        ...rest,
+        connections: rest.connections ?? (connection ? [connection] : undefined),
         type: 'webPubSubTrigger',
     });
 }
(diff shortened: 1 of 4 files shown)


types/webpubsub.d.ts
```diff
@@ -94,12 +93,11 @@ export interface WebPubSubContextInputOptions {
...
 
     /**
-     * Optional - The name of an app settings or setting collection that specifies the upstream Azure Web PubSub service.
+     * Optional - The names of app settings or setting collections that specify the upstream Azure Web PubSub services.
      * The value is used for Abuse Protection and Signature validation.
      * The value is auto resolved with "WebPubSubConnectionString" by default.
-     * Null means the validation isn't needed and always succeeds.
      */
-    connection?: string;
+    connections?: string[];
```
swapnil-nagar, line 107: This change is not backward compatible and introduces a breaking change for the customers.

Has the contract for Web PubSub changed? The extension should support both argument formats, with the connection parameter marked as deprecated.

src/trigger.ts
```diff
@@ -129,8 +129,11 @@ export function mySql(options: MySqlTriggerOptions): MySqlTrigger {
 }
 
 export function webPubSub(options: WebPubSubTriggerOptions): WebPubSubTrigger {
+    // eslint-disable-next-line deprecation/deprecation
+    const { connection, ...rest } = options;
     return addTriggerBindingName({
-        ...options,
+        ...rest,
+        connections: rest.connections ?? (connection ? [connection] : undefined),
```
swapnil-nagar, line 138: Can we have unit test coverage for the change?

swapnil-nagar (changes requested):

src/trigger.ts
Y-Sindo: Thanks for the review @swapnil-nagar. I have added some unit tests.

JialinXin (approved):

swapnil-nagar (approved):

PR #434, opened 2026-05-13
Adding the generic connector Trigger

swapnil-nagar:
Adds native connectorTrigger, connectorContent input, and connectorContent output binding support to @azure/functions, enabling Azure Logic Apps connector events (Office 365, SharePoint, Teams, Kusto, etc.) to be registered as first-class function triggers and bindings on par with existing bindings like Cosmos DB, Event Hub, and Service Bus.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -128,6 +128,10 @@ export function eventGrid(name: string, options: EventGridFunctionOptions): void
     generic(name, convertToGenericOptions(options, trigger.eventGrid));
 }
 
+export function connectorTrigger<T = unknown>(name: string, options: ConnectorTriggerFunctionOptions<T>): void {
+    generic(name, convertToGenericOptions(options, trigger.connectorTrigger));
+}
+
 export function cosmosDB(name: string, options: CosmosDBFunctionOptions): void {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
     generic(name, convertToGenericOptions(options, <any>trigger.cosmosDB));
@@ -180,10 +184,6 @@ export function mcpResource(name: string, options: McpResourceFunctionOptions): 
  */
 export function mcpPrompt(name: string, options: McpPromptFunctionOptions): void {
     generic(name, convertToGenericOptions(options, trigger.mcpPrompt));
-}
-
-export function connectorTrigger(name: string, options: ConnectorTriggerFunctionOptions): void {
-    generic(name, convertToGenericOptions(options, trigger.connectorTrigger));
 }
 
 export function generic(name: string, options: GenericFunctionOptions): void {
--- a/src/app.ts
+++ b/src/app.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 import {
+    ConnectorTriggerFunctionOptions,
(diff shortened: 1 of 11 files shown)


TsuyoshiUshio:
## Review — Adding the generic connector Trigger

Thanks for the PR! The structure follows existing binding patterns well. A few items to consider:

### Must-fix

**1. ConnectorTriggerHandler should have a generic type parameter**

Currently:
```typescript
export type ConnectorTriggerHandler = (triggerInput: unknown, context: InvocationContext) => FunctionResult;
```

Should be (consistent with McpToolTriggerHandler<T>):
```typescript
export type ConnectorTriggerHandler<T = unknown> = (triggerInput: T, context: InvocationContext) => FunctionResult;
```

Without <T>, users must cast every trigger input. Adding <T> later would be a breaking change since ConnectorTriggerFunctionOptions also needs to propagate it.

### Should-fix

**2. No tests for pp.connectorTrigger()**

	rigger.connectorTrigger(), input.connectorContent(), and output.connectorContent() are individually tested — but the main entry point pp.connectorTrigger() (which calls convertToGenericOptions + generic) is not covered. Other bindings have corresponding tests in 	est/app.test.ts.

**3. JSDoc is minimal**

	ypes/input.d.ts and 	ypes/output.d.ts have one-line docs for connectorContent(). Compare with existing bindings like sql() or cosmosDB() which include @param annotations. Same for 	ypes/trigger.d.ts.

**4. Alphabetical ordering in src/app.ts**

connectorTrigger() is placed right before generic(). Existing bindings are in roughly alphabetical order — it should go after cosmosDB() to stay consistent.

### Nice-to-have

- Consider renaming the unused handler variable in tests to _handler to avoid the slint-disable-next-line comment.
- Branch name typo: genericCOnnector

### Summary

Good first-class binding addition that follows the established patterns closely. The generic type parameter is the most important item — it's an API contract decision that's hard to change after release. Everything else is polish.

TsuyoshiUshio:
## Re-review — All feedback addressed

Thanks @swapnil-nagar! Checked the second commit and all items from my initial review are resolved:

| Item | Status |
|------|--------|
| Generic type parameter `<T>` on `ConnectorTriggerHandler` | Fixed — now `ConnectorTriggerHandler<T = unknown>` with propagation to `ConnectorTriggerFunctionOptions<T>` |
| `app.connectorTrigger()` tests missing | Fixed — 2 tests added (basic registration + with extraInputs/extraOutputs) |
| JSDoc incomplete | Fixed — `@param` annotations added to `types/input.d.ts`, `types/output.d.ts`, `types/trigger.d.ts` |
| Alphabetical ordering in `src/app.ts` | Fixed — placed after `eventGrid()` and before `cosmosDB()` |
| `eslint-disable` for unused handler | Fixed — renamed to `_handler` / `_triggerInput` / `_context` |

LGTM from a Node.js SDK perspective. One remaining question from my earlier review still stands: is the corresponding .NET extension (`connectorTrigger` / `connectorContent` binding types) already available or in-progress? Without it the Host will reject the binding type at runtime. Would be good to mention the extension status in the PR description for reviewers.

TsuyoshiUshio (approved):

PR #438, opened 2026-05-15
Simplify connectorTrigger: remove unreleased input/output bindings

swapnil-nagar:
### **Summary**
Removes the connectorContent input and output binding support from the connectorTrigger feature and simplifies ConnectorTriggerOptions to an open-ended Record<string, unknown> (no required properties).

### Motivation

The connectorContent input/output bindings (input.connectorContent(), output.connectorContent()) and the connection / connector / triggerOperation properties on ConnectorTriggerOptions were added speculatively in #434. From the connectors team perspective:

- The input/output bindings are intentionally not exposed yet because the customer-facing parameter shape is not finalized.
- The C# host extension (ConnectorTriggerAttribute) is empty — it is a pure webhook receiver that routes by function name. The connection, connector, and triggerOperation properties were metadata-only for trigger config registration and are not consumed by the runtime.
- There is no backward-compatibility concern because this package version is still in preview.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@azure/functions",
-    "version": "4.14.0",
+    "version": "4.15.0-preview",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@azure/functions",
-            "version": "4.14.0",
+            "version": "4.15.0-preview",
             "license": "MIT",
             "dependencies": {
                 "@azure/functions-extensions-base": "0.2.0",
@@ -61,7 +61,6 @@
             "version": "0.2.0",
             "resolved": "https://registry.npmjs.org/@azure/functions-extensions-base/-/functions-extensions-base-0.2.0.tgz",
             "integrity": "sha512-ncCkHBNQYJa93dBIh+toH0v1iSgCzSo9tr94s6SMBe7DPWREkaWh8cq33A5P4rPSFX1g5W+3SPvIzDr/6/VOWQ==",
-            "license": "MIT",
             "engines": {
                 "node": ">=18.0"
             }
@@ -71,7 +70,6 @@
             "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz",
             "integrity": "sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw==",
             "dev": true,
-            "license": "MIT",
             "dependencies": {
                 "@babel/highlight": "^7.10.4"
             }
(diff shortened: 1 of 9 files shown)


TsuyoshiUshio:
Thanks for the update. I reviewed the PR and the direction looks good to me.

Given the Connectors team feedback that connector input/output bindings should not be exposed to customers yet, and since this package version has not been released, removing `input.connectorContent(...)` / `output.connectorContent(...)` and the related types is reasonable. It also keeps the preview surface smaller until the required parameter shape is finalized.

I validated the PR locally:

- `npm ci` succeeded
- `npm test -- --runInBand` succeeded: 362 passing
- `npm run lint` succeeded with no changes
- `npm run build` succeeded
- `npm audit --omit=dev` reports 0 production vulnerabilities

A couple of non-blocking suggestions before merge:

1. The test named `should generate different names for different options` now passes two empty option objects and expects the names to be equal. Consider renaming it to something like `should generate a stable binding name for equivalent options`.
2. The `package-lock.json` diff updates a number of transitive dependencies even though `package.json` only bumps the package version. If that dependency refresh is intentional, it would be helpful to mention it in the PR description; otherwise, consider minimizing the lockfile diff to reduce review noise.
3. Please consider adding the Connectors team rationale to the PR description, e.g. that connector input/output bindings are intentionally not exposed yet because the customer-facing parameter shape is not finalized, and there is no backward compatibility concern because this package version has not been released.

No blocking issue found from my side.

swapnil-nagar:
> 1. The test named `should generate different names for different options` now passes two empty option objects and expects the names to be equal. Consider renaming it to something like `should generate a stable binding name for equivalent options`.

Renamed the test.

Added PR Description for 2 & 3.

TsuyoshiUshio:
Thanks for the updates. I re-reviewed the latest PR head (`abe6221`) and the changes look good from the code side.

Validated again locally:

- `npm test -- --runInBand` succeeded: 362 passing
- `npm run lint` succeeded with no changes
- `npm run build` succeeded
- `npm audit --omit=dev` reports 0 production vulnerabilities

Also checked the PR checks: all visible GitHub / Azure Pipelines checks are passing.

The renamed test looks good, and the updated PR description now captures the Connectors team rationale and the no-backward-compatibility point.

One small non-blocking nit: the PR description currently has several `vscode-file://...` links around terms like `connectorContent`, `ConnectorTriggerOptions`, and `Record<string, unknown>`. Those links will not work for GitHub readers, so I would suggest changing them to inline code formatting instead.

No blocking issue found from my side.

TsuyoshiUshio (approved):

swapnil-nagar:
> Thanks for the updates. I re-reviewed the latest PR head (`abe6221`) and the changes look good from the code side.
> 
> Validated again locally:
> 
> * `npm test -- --runInBand` succeeded: 362 passing
> * `npm run lint` succeeded with no changes
> * `npm run build` succeeded
> * `npm audit --omit=dev` reports 0 production vulnerabilities
> 
> Also checked the PR checks: all visible GitHub / Azure Pipelines checks are passing.
> 
> The renamed test looks good, and the updated PR description now captures the Connectors team rationale and the no-backward-compatibility point.
> 
> One small non-blocking nit: the PR description currently has several `vscode-file://...` links around terms like `connectorContent`, `ConnectorTriggerOptions`, and `Record<string, unknown>`. Those links will not work for GitHub readers, so I would suggest changing them to inline code formatting instead.
> 
> No blocking issue found from my side.

Removed hyperlinks from PR Descriptions

PR #440, opened 2026-05-19
Harden HTTP proxy header handling

TsuyoshiUshio:
## Summary
- block hop-by-hop and framing response headers from the HTTP proxy response path
- handle invalid `x-ms-client-principal` payloads without failing the invocation
- add focused tests for proxy header filtering and invalid client principal JSON

## Notes
F4 (`ensureErrorType` object serialization) is intentionally not changed in this PR. That area needs a separate design discussion because the safer fix likely requires sanitizer behavior aligned with azure-functions-host and azure-functions-nodejs-worker, not just using an object's `message` property.

## Validation
- `npm test` (375 passed)
- `npm run lint`

### DIFF
--- a/src/http/extractHttpUserFromHeaders.ts
+++ b/src/http/extractHttpUserFromHeaders.ts
@@ -3,6 +3,7 @@
 
 import { HttpRequestUser } from '@azure/functions';
 import { nonNullValue } from '../utils/nonNull';
+import { workerSystemLog } from '../utils/workerSystemLog';
 
 /* grandfathered in. Should fix when possible */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access */
@@ -12,7 +13,16 @@ export function extractHttpUserFromHeaders(headers: Headers): HttpRequestUser | 
 
     const clientPrincipal = headers.get('x-ms-client-principal');
     if (clientPrincipal) {
-        const claimsPrincipalData = JSON.parse(Buffer.from(clientPrincipal, 'base64').toString('utf-8'));
+        let claimsPrincipalData: any;
+        try {
+            claimsPrincipalData = JSON.parse(Buffer.from(clientPrincipal, 'base64').toString('utf-8'));
+        } catch (err) {
+            workerSystemLog(
+                'warning',
+                `Failed to parse x-ms-client-principal header: ${err instanceof Error ? err.message : String(err)}`
+            );
+            return null;
+        }
 
         if (claimsPrincipalData['identityProvider']) {
             user = {
--- a/src/http/extractHttpUserFromHeaders.ts
+++ b/src/http/extractHttpUserFromHeaders.ts
@@ -24,6 +24,15 @@ export function extractHttpUserFromHeaders(headers: Headers): HttpRequestUser | 
             return null;
         }
 
+        if (
+            claimsPrincipalData === null ||
+            typeof claimsPrincipalData !== 'object' ||
+            Array.isArray(claimsPrincipalData)
+        ) {
+            workerSystemLog('warning', 'Parsed x-ms-client-principal header was not a JSON object.');
+            return null;
+        }
+
         if (claimsPrincipalData['identityProvider']) {
             user = {
                 type: 'StaticWebApps',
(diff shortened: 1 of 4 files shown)


src/http/extractHttpUserFromHeaders.ts
```diff
@@ -12,7 +13,16 @@ export function extractHttpUserFromHeaders(headers: Headers): HttpRequestUser |
 
     const clientPrincipal = headers.get('x-ms-client-principal');
     if (clientPrincipal) {
-        const claimsPrincipalData = JSON.parse(Buffer.from(clientPrincipal, 'base64').toString('utf-8'));
+        let claimsPrincipalData: any;
+        try {
```
swapnil-nagar, line 17: The new try/catch handles invalid JSON syntax which is great, but this code can still throw for valid JSON that is not an object.


I think we should also validate that `claimsPrincipalData` is a non-null object before reading from it, and add a test for inputs like `null` or `"string"`.

swapnil-nagar: [Please check] Earlier extractHttpUserFromHeaders would throw exception when JSON parsing fails, but now now behavior is changes to return null, see if silently would that cause any downstream issues?

src/http/httpProxy.ts
```diff
@@ -40,7 +53,9 @@ export async function sendProxyResponse(invocationId: string, userRes: HttpRespo
     const proxyRes = nonNullProp(responses, invocationId);
     delete responses[invocationId];
     for (const [key, val] of userRes.headers.entries()) {
-        proxyRes.setHeader(key, val);
+        if (isAllowedProxyResponseHeader(key)) {
+            proxyRes.setHeader(key, val);
+        }
```
swapnil-nagar, line 58: `Content-Length` is an end-to-end header, and customers may set it intentionally for HEAD/304/206 responses, downloads, or other scenarios where clients rely on it. Dropping it could change behavior for existing apps.

Can you check RFC Compliant docs if we can preseves 'content-length'.

swapnil-nagar (changes requested):

src/http/extractHttpUserFromHeaders.ts
TsuyoshiUshio: Added validation that the parsed \x-ms-client-principal\ value is a non-null JSON object, and added coverage for \
ull\ and string inputs. I also reject arrays since they are not a valid claims object shape.

TsuyoshiUshio: I checked the downstream behavior. Returning \
ull\ follows the existing 'no user principal' path and avoids failing the invocation for a malformed auth header. We still log a warning so the malformed header is diagnosable.

src/http/httpProxy.ts
TsuyoshiUshio: Good point. \Content-Length\ is an end-to-end header, and dropping it may have compatibility impact for HEAD/304/206/download scenarios. I am leaving this thread open for discussion rather than changing it blindly. My original concern was mismatched framing when proxying streamed bodies; we should decide whether preserving it is preferred or whether only conflicting cases should be filtered.

swapnil-nagar: So what's your recommendation?

TsuyoshiUshio: My recommendation is to preserve `Content-Length` and remove it from the blocked header list.

`Content-Length` is end-to-end response metadata, not a hop-by-hop header, and removing other hop-by-hop headers does not change the response body bytes. Dropping it could break existing apps that intentionally set it, especially for HEAD/304/206/download scenarios.

The original concern was about forwarding connection/framing controls through this proxy. This code copies user response headers to Node.js `ServerResponse`, then writes the response body through `proxyRes.write(...)`. If we forward headers like `Connection`, `TE`, `Trailer`, `Transfer-Encoding`, or `Upgrade`, those headers can describe connection-specific behavior or wire framing that should be owned by the proxy/Node response layer instead.

To address that without treating `Content-Length` as hop-by-hop, I updated the implementation to:
- allow `content-length`;
- keep blocking the standard hop-by-hop headers;
- also strip any header names listed by the `Connection` header, because `Connection: x-private-hop` marks `x-private-hop` as connection-specific and a proxy should not forward it.

I am not validating `Content-Length` against the streamed body here, because that would require buffering/counting the stream and would weaken the streaming path. If a user sets an incorrect `Content-Length`, that is a response correctness issue separate from this hop-by-hop proxy hardening.

swapnil-nagar (approved):

PR #441, opened 2026-05-19
Sanitize serialized library errors

TsuyoshiUshio:
## Summary
- add credential-aware sanitization when serializing non-Error thrown values
- redact credential-like property names and known credential tokens while preserving useful object context
- handle circular references in thrown objects without throwing during error conversion
- add tests for credential properties, credential tokens, URL credentials, and circular references

## Notes
This is a first implementation intended to support review and discussion. The worker PR handles the primary invocation-failure path before errors are sent to the host; this library PR keeps `@azure/functions` behavior aligned for library-level conversions/logging.

## Validation
- `npm test` (378 passing)
- `npm run lint`


src/errors.ts
```diff
@@ -34,6 +34,29 @@ export class ReadOnlyError extends AzFuncTypeError {
...
+    '&code=',
+    '?code=',
+    '/code=',
+    'key=',
+];
+const urlCredentialPattern = /\b([a-zA-Z]+):\/\/([^:/\s]+):([^@/\s]+)@([^:/\s]+):([0-9]+)\b/g;
+
 export function ensureErrorType(err: unknown): ValidatedError {
     if (err instanceof Error) {
         return err;
```
swapnil-nagar, line 41: Do you need to Sanitize the Error instance as well?

src/errors.ts
```diff
@@ -42,16 +65,84 @@ export function ensureErrorType(err: unknown): ValidatedError {
...
+            message = safeStringify(err);
         } else {
-            message = String(err);
+            message = sanitizeErrorString(String(err));
         }
         return new Error(message);
     }
 }
 
+export function sanitizeErrorString(input: string): string {
```
swapnil-nagar: Sanitize string and checking credential should be out of the errors.ts. Also are there other places we should be applying this changes?

src/errors.ts
```diff
@@ -34,6 +34,29 @@ export class ReadOnlyError extends AzFuncTypeError {
     }
 }
 
+const hiddenCredential = '[Hidden Credential]';
```
swapnil-nagar: https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/src/http/httpProxy.ts
At 107 ensureErrorType is checked but raw erro.stack and err.message is logged. 

You may want to change that logic as well.

swapnil-nagar: Same for https://github.com/Azure/azure-functions-nodejs-library/blob/v4.x/src/hooks/registerHook.ts at line 67.

src/errors.ts
```diff
@@ -42,16 +65,84 @@ export function ensureErrorType(err: unknown): ValidatedError {
...
+    });
+}
+
+function isCredentialName(name: string): boolean {
+    return credentialNameFragments.some((fragment) => name.toLowerCase().includes(fragment));
+}
+
 export function trySetErrorMessage(err: Error, message: string): void {
     try {
         err.message = message;
```
swapnil-nagar: You need to sanitize the err at here as well.

swapnil-nagar (approved):

swapnil-nagar (changes requested):
Reverting accidental approavl

src/errors.ts
TsuyoshiUshio: Yes. I kept \nsureErrorType\ from mutating existing Error instances directly, and instead sanitize at the message/logging boundaries. \	rySetErrorMessage\ now sanitizes before assigning a message.

src/errors.ts
TsuyoshiUshio: Moved the sanitizer into \src/utils/errorSanitizer.ts\. I also applied it to the HTTP proxy server error log and the \	rySetErrorMessage\ path.

src/errors.ts
TsuyoshiUshio: Updated the HTTP proxy server error log to sanitize the stack/message before writing the worker system log.

TsuyoshiUshio: I checked \
egisterHook.ts\. That path rethrows the original error and does not log or serialize it there, so I left it unchanged to avoid mutating what the caller catches. The worker serialization/logging boundary now sanitizes the error text.

src/errors.ts
TsuyoshiUshio: Updated this path by sanitizing in \	rySetErrorMessage\, with a test covering credential redaction.

src/utils/credentialSanitizer.ts
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License.
+
```
swapnil-nagar, line 3: NIT: You can rename the file to credentialSanitizer.ts since it is only sanitizing strings, regardless of error or not.

swapnil-nagar (dismissed):

src/utils/credentialSanitizer.ts
TsuyoshiUshio: Renamed to `credentialSanitizer.ts` and updated the imports in `errors.ts`, `httpProxy.ts`, and `errors.test.ts`.

swapnil-nagar (approved):

Issue #444, opened 2026-06-05
McpToolResponse.isError is dropped before MCP tool result reaches the client


codylittle:
`McpToolResponse` accepts `isError: true`, but the Node.js worker conversion path appears to drop that field before the result is sent to the Azure Functions MCP extension. As a result, MCP clients receive a
successful `tools/call` result instead of a tool-error result.

## Environment

- `@azure/functions`: `4.16.0`
- Azure Functions MCP extension: observed with extension bundle preview using MCP extension `1.5.0`
- Runtime: Node.js Azure Functions app using `app.mcpTool(...)`

## Reproduction

Return an MCP tool response with `isError: true`:

```ts
return new McpToolResponse({
  content: [
    new McpTextContent("Invalid file")
  ],
  isError: true,
});
```

The public type supports this:

```ts
export interface McpToolResponseInit {
  content: McpContentBlock[];
  structuredContent?: unknown;
  isError?: boolean;
}
```

But toMcpToolResult() serializes only type, content, and optional structuredContent:

```ts
const out: McpToolResult = { type, content: contentStr };

if (response.structuredContent !== undefined && response.structuredContent !== null) {
  out.structuredContent =
    typeof response.structuredContent === "string"
      ? response.structuredContent
      : JSON.stringify(response.structuredContent);
}

return out;
```

## Actual Behavior

A handler-level error response reaches the MCP client like this:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Invalid file"
      }
    ]
  },
  "id": 5,
  "jsonrpc": "2.0"
}
```

There is no isError field.

## Expected Behavior

The client-visible MCP result should include:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Invalid file"
      }
    ],
    "isError": true
  },
  "id": 5,
  "jsonrpc": "2.0"
}
```

Per the MCP schema, missing isError defaults to false, so clients currently treat handler-level tool errors as successful tool results.

Spec reference: https://modelcontextprotocol.io/specification/2025-06-18/schema#tools-call

# Notes

I'm unsure if this would best be solved in the node library, or in the extension handler same way `_meta` is (in preview builds)

lilyjma:
thanks for reporting this @codylittle!  We talked about this in our status update today. @swapnil-nagar will look into this and address your question

swapnil-nagar:
[...]

PR #446, opened 2026-06-15
isError Pass Through with the MCP Tool Call

swapnil-nagar:
https://github.com/Azure/azure-functions-nodejs-library/issues/444

### DIFF
--- a/src/converters/toMcpToolResult.ts
+++ b/src/converters/toMcpToolResult.ts
@@ -8,6 +8,7 @@ import { shouldCreateStructuredContentMarker } from '../utils/mcpContentMarker';
 
 const multiContentResultType = 'multi_content_result';
 const textContentResultType = 'text';
+const callToolResultType = 'call_tool_result';
 
 /**
  * Converts a tool handler's return value into the wire-format MCP tool result.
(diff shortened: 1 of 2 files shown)


src/converters/toMcpToolResult.ts
```diff
@@ -63,6 +64,16 @@ export function toMcpToolResult(result: unknown, context?: InvocationContext): M
 function serializeToolResponse(response: McpToolResponse): McpToolResult {
     const blocks = ensureTextBlockWhenStructured(response);
 
+    // When `isError` is set, route the response through the host's `call_tool_result`
+    // passthrough so the boolean reaches the MCP client. The host's ToolReturnValueBinder
+    // recognizes `type == "call_tool_result"` and deserializes `content` directly into
+    // the MCP SDK's full `CallToolResult` (including `isError`, `structuredContent`, `_meta`).
+    // The non-passthrough envelope only carries `content` + `structuredContent`, so any
+    // `isError` value would otherwise be silently dropped.
+    if (response.isError !== undefined) {
```
liliankasem, line 73: Honestly there is not reason not to do this for all of the responses; instead of an if-statement here, why not just use CallToolResult for everything?

PR #454, opened 2026-06-19
CosmosDb ChangeFeed Support

swapnil-nagar:
Adds Cosmos DB Change Feed mode support and validation so changeFeedMode is surfaced in types and preserved through function registration metadata.

## Changes
- Added CosmosDBv4ChangeFeedMode type with:
  - LatestVersion
  - AllVersionsAndDeletes
- Added changeFeedMode to Cosmos DB v4 trigger options.
- Re-exported CosmosDBChangeFeedMode from shared Cosmos DB types.
- Updated Cosmos DB API docs to clarify default behavior and how to enable full change feed processing.
- Added tests for:
  - Metadata conversion path (toCoreFunctionMetadata)
  - End-to-end app registration path (app.cosmosDB)

## Version
- Bumped package version to 4.16.1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@azure/functions",
-    "version": "4.16.0",
+    "version": "4.16.1",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@azure/functions",
-            "version": "4.16.0",
+            "version": "4.16.1",
             "license": "MIT",
             "dependencies": {
                 "@azure/functions-extensions-base": "0.3.0",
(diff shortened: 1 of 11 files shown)


jviau:
I am not familiar with E2E-test situation for this repo, but want to make sure the feature is validated either manually or by tests. A specific scenario to validate for is the incoming payload type when using `AllVersionsAndDelete`. With `LatestVersion` it will be `T`, for AVAD it will be `ChangeFeedItem<T>`. So the JSON representation of this CSharp class: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.changefeeditem-1?view=azure-dotnet-preview

TsuyoshiUshio:
+1 for the Jacob's comment. Change looks good. E2E might be required. 
For E2E testing with this change, If you've already done the manual E2E, I'd happy to approve. But I'm wondering if we can write E2E testing for nodejs-library/worker/extensions if one of them are changed, We can use https://github.com/Azure/azure-functions-nodejs-e2e-tests .

package-lock.json
```diff
@@ -2881,20 +2881,6 @@
             "integrity": "sha512-QMUezzXWII9EV5aTFXW1UBVUO77wYPpjqIF8/AviUCThNeSYZykpoTixUeaNNBwmCev0AMDWMAni+f8Hxb1IFw==",
             "dev": true
         },
-        "node_modules/fsevents": {
```
TsuyoshiUshio: nit: This is not blocking issue.  This library is dev dependency for MacOS. No harm but it is reviewing noise. :)

swapnil-nagar:
> +1 for the Jacob's comment. Change looks good. E2E might be required. For E2E testing with this change, If you've already done the manual E2E, I'd happy to approve. But I'm wondering if we can write E2E testing for nodejs-library/worker/extensions if one of them are changed, We can use https://github.com/Azure/azure-functions-nodejs-e2e-tests .

Emulator do not support this. I will be creating a new sample for this feature and test locally

swapnil-nagar:
@copilot resolve the merge conflicts in this pull request

jviau (approved):
Curious where `FullFidelity` came from? Does the node CosmosDB package use that terminology?

swapnil-nagar:
> Curious where `FullFidelity` came from? Does the node CosmosDB package use that terminology?

@jviau No, I did not like the name All Version and Delete and fullFidelity sounds next best option so changed to that. I can revert if that's the name we plan to go.

TsuyoshiUshio (approved):

PR #461, opened 2026-07-27
Fix GET/HEAD request bodies failing before the handler

### LINKED ISSUE #458: RPC HttpRequest passes GET/HEAD bodies to WHATWG Request and fails before handler
### Version

- `@azure/functions`: reproduced on 4.16.0, 4.16.1, and 4.16.2
- Node.js: production Node 20; also reproduced locally on Node 24
- Azure Functions Core Tools: 4.8.0
- Hosting: Azure Static Web Apps managed Functions

### Reproduction

```js
const { HttpRequest } = require('@azure/functions');

new HttpRequest({
  url: 'https://example.test/api/probe',
  method: 'GET',
  body: { string: 'x' },
});
```

This throws before an application handler can run:

```text
TypeError: Request with GET/HEAD method cannot have body.
    at new Request (node:internal/deps/undici/undici:10023:17)
    at new HttpRequest (@azure/functions/dist/azure-functions.js:2136:25)
    at fromRpcTypedData
    at InvocationModel.getArguments
```

The public SWA behavior was also deterministic before applying an application workaround:

```text
curl -X GET --data x https://red-river-0f7efed00.7.azurestaticapps.net/api/health/bootstrap
```

A GET with a body returned a bare 500 and never reached the handler. The same request with no body or `Content-Length: 0` reached the handler and returned its expected 401 JSON auth response.

### Cause

In 4.16.x, the non-streaming RPC `HttpRequest` constructor passes the decoded RPC body to the global WHATWG `Request` for every method. The streaming `createStreamRequest` path already omits the body for GET and HEAD. The two paths therefore behave differently.

Organic body provenance is unknown because the available telemetry does not retain request headers/body length or client identity. Possible sources include malformed external traffic or an upstream host/proxy handoff; this report only claims the deterministic library failure once RPC data contains a GET/HEAD body.

### Expected behavior

The RPC path should match the streaming path and omit `body` for GET and HEAD before constructing the WHATWG `Request`. A regression test should cover GET/HEAD bodies (including zero-length RPC bodies), mixed-case methods if accepted internally, and preservation of POST/PUT/PATCH bodies.

swapnil-nagar:
Issue https://github.com/Azure/azure-functions-nodejs-library/issues/458

Aligns the HttpRequest constructor with the existing streaming path by omitting the request body for GET/HEAD methods, so the worker no longer crashes constructing Node's global WHATWG Request (which rejects bodies on GET/HEAD). Fixes #458. Also omits the response body for null-body statuses (204/205/304) in HttpResponse to avoid the equivalent Response TypeError. Adds unit tests covering GET/HEAD body handling (incl. mixed-case methods) and null-body status responses.

### DIFF
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -33,17 +33,24 @@ export class HttpRequest implements types.HttpRequest {
         let nativeReq = init.nativeRequest;
         if (!nativeReq) {
             const url = nonNullProp(init, 'url');
+            const method = nonNullProp(init, 'method');
 
+            // GET/HEAD requests cannot have a body. The global Request constructor throws if one is
+            // provided, so omit it for these methods to match the streaming path (createStreamRequest).
+            // See https://github.com/Azure/azure-functions-nodejs-library/issues/458
             let body: Buffer | string | undefined;
-            if (init.body?.bytes) {
-                body = Buffer.from(init.body?.bytes);
-            } else if (init.body?.string) {
-                body = init.body.string;
+            const lowerMethod = method.toLowerCase();
+            if (lowerMethod !== 'get' && lowerMethod !== 'head') {
+                if (init.body?.bytes) {
+                    body = Buffer.from(init.body?.bytes);
+                } else if (init.body?.string) {
+                    body = init.body.string;
+                }
             }
 
             nativeReq = new Request(url, {
                 body,
-                method: nonNullProp(init, 'method'),
+                method,
                 headers: fromNullableMapping(init.nullableHeaders, init.headers),
             });
         }
--- a/src/http/HttpRequest.ts
+++ b/src/http/HttpRequest.ts
@@ -13,6 +13,7 @@ import { fromNullableMapping } from '../converters/fromRpcNullable';
 import { fromRpcTypedData } from '../converters/fromRpcTypedData';
 import { AzFuncSystemError } from '../errors';
 import { isDefined, nonNullProp } from '../utils/nonNull';
+import { workerSystemLog } from '../utils/workerSystemLog';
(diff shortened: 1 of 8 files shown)


test/converters/toRpcHttp.test.ts
```diff
@@ -95,6 +95,18 @@ describe('toRpcHttp', () => {
         await expect(toRpcHttp('invocId', { status: {} })).to.eventually.be.rejectedWith(/status/i);
     });
 
+    // https://github.com/Azure/azure-functions-nodejs-library/issues/458 (Response-side parallel):
+    // null-body statuses must not crash conversion when a handler returns a body.
```
hallvictoria, line 99: nit: add test for 205 here as well

src/http/HttpRequest.ts
```diff
@@ -33,17 +33,24 @@ export class HttpRequest implements types.HttpRequest {
...
+            // GET/HEAD requests cannot have a body. The global Request constructor throws if one is
+            // provided, so omit it for these methods to match the streaming path (createStreamRequest).
+            // See https://github.com/Azure/azure-functions-nodejs-library/issues/458
             let body: Buffer | string | undefined;
-            if (init.body?.bytes) {
-                body = Buffer.from(init.body?.bytes);
-            } else if (init.body?.string) {
-                body = init.body.string;
+            const lowerMethod = method.toLowerCase();
+            if (lowerMethod !== 'get' && lowerMethod !== 'head') {
```
hallvictoria, line 44: Would it be helpful to add a log when the method is GET / HEAD saying that the body is discarded?

This is the correct behavior to avoid crashing, but right now the body is discarded silently, which might make debugging difficult

swapnil-nagar: Added a log

hallvictoria (approved):

PR #474, opened 2026-09-16
AllVersionsAndDeletes Name Change

swapnil-nagar:
## Summary

Updates Cosmos DB change feed mode support to use the host-aligned `AllVersionsAndDeletes` contract and adds sample coverage for validating both change feed modes.

## Changes

- Renamed the full-fidelity change feed mode surface to `AllVersionsAndDeletes`.
- Added enum-backed change feed mode contracts:
  - `LatestVersion`
  - `AllVersionsAndDeletes`
- Updated Cosmos DB v4 trigger typings so `changeFeedMode` uses the enum contract.
- Updated `app.cosmosDB(...)` registration coverage for the `AllVersionsAndDeletes` path.
- Updated samples to demonstrate:
  - `LatestVersion` with plain document payloads
  - `AllVersionsAndDeletes` with `CosmosDBChangeFeedItem<T>` payloads
- Configured the sample app for the test Cosmos DB account/container.

## Validation

- Ran TypeScript validation successfully:

```powershell
npx tsc -p . --noEmit

### CODE REVIEW

File: src/trigger.ts [hunk 1]
```diff
@@ -108,12 +108,7 @@ export function cosmosDB(options: CosmosDBTriggerOptions): CosmosDBTrigger {
         type: 'cosmosDBTrigger',
     });
 
-    // The Functions host (Cosmos DB extension) only recognizes 'AllVersionsAndDeletes' on the wire,
-    // so translate the developer-facing 'FullFidelity' value before registration.
-    if ('changeFeedMode' in binding && binding.changeFeedMode === 'FullFidelity') {
```
TsuyoshiUshio, line 113: Removing `FullFidelity` is breaking change. If we go with it, We should make it major version up. (not patch)
We should convert `FullFidelity` to `AllVersionsAndDeletes` if customers are specified it.

types/cosmosDB.v4.d.ts
```diff
@@ -33,7 +33,10 @@ export interface CosmosDBv4ChangeFeedItem<T = unknown> {
 /**
  * Change feed mode for Cosmos DB trigger bindings.
  */
-export type CosmosDBv4ChangeFeedMode = 'LatestVersion' | 'FullFidelity';
+export const enum CosmosDBv4ChangeFeedMode {
```
TsuyoshiUshio: Please do not export an ambient `const enum`. It causes TS2748 when a customer uses `isolatedModules`. Please use a regular `enum` or keep the string union.

types/app.d.ts
```diff
@@ -155,7 +155,7 @@ export function eventGrid<T = EventGridEvent>(name: string, options: EventGridFu
 /**
  * Registers a Cosmos DB function in your app that will be triggered by container changes.
  * By default, this trigger processes latest document versions. To include delete events and
- * intermediate mutations, set `changeFeedMode` to `FullFidelity`.
+ * intermediate mutations, set `changeFeedMode` to `FullFidelity` (host feature name: `AllVersionsAndDeletes`).
```
TsuyoshiUshio: This text tells customers to use `FullFidelity`, but the new type does not accept this value. Please change the text to `AllVersionsAndDeletes`.

harshivcodes: Should this recommend `CosmosDBv4ChangeFeedMode.AllVersionsAndDeletes` instead of `FullFidelity`?

harshivcodes: Tsuyoshi also called this out, fyi

src/trigger.ts
harshivcodes: +1 If we are releasing a patch version with this change, could we retain the translation for one version (mark it as deprecated) and then remove it in a later version.

Alternatively, major version up would also work if we don't want to do multiple releases.

swapnil-nagar: This is not a breaking change as library version for this feature is not released yet. And I do  not see any usage as well. As this feature is in preview.

types/cosmosDB.v4.d.ts
swapnil-nagar: Good point, fixed and updated the tsconfig to avoid future changes :)

types/app.d.ts
swapnil-nagar: Done, fixed

swapnil-nagar: Fixed

harshivcodes (dismissed):

types/cosmosDB.v4.d.ts
```diff
@@ -51,14 +54,14 @@ export interface CosmosDBv4LatestVersionFunctionOptions<T = unknown>
     retry?: RetryOptions;
 }
 
-export interface CosmosDBv4FullFidelityFunctionOptions<T = unknown>
+export interface CosmosDBv4AllVersionsAndDeletesFunctionOptions<T = unknown>
```
TsuyoshiUshio, line 57: `CosmosDBv4FullFidelityFunctionOptions` is public in v4.16.4. This rename removes a public type in a patch release. Please keep the old name as a deprecated type alias.

src/trigger.ts
TsuyoshiUshio: @azure/functions 4.16.4 is published on npm. The v4.16.4 tag contains commit `00e0a890`. Customers can already use `FullFidelity`. Please keep the translation or release a major version.

types/cosmosDB.v4.d.ts
swapnil-nagar: Addressed this comment

src/trigger.ts
TsuyoshiUshio: This breaking change is still present. Please keep the `FullFidelity` to `AllVersionsAndDeletes` translation for existing customers.

TsuyoshiUshio (approved):

